PR #8, opened 2023-05-19
add scripts to fix (spatial) restart problem

keiyamamo:
Hi @dbruneau-mie 

This PR adds two scripts that can be used to fix h5/xdmf files with inconsistent mesh information. 
When restarting turtleFSI, h5 files can contain mesh/vector information with different ordering due to different mesh partitioning and `check_h5file.py` can be used to correct the mesh/vector ordering based on one of the h5 files. 
After fixing, one can use `combine_xdmf.py’ to safely merge xdmf files. This also means that it is better not to merge visualization files at the end of the simulation, but we can come back to this in the turtle meeting.

Please let me know if you have any questions!

Best,
Kei

### DIFF
--- a/postprocessing_h5py/check_h5file.py
+++ b/postprocessing_h5py/check_h5file.py
@@ -62,10 +62,13 @@ def main(folder, correct, wrong):
     map_index = []
     
     # create a list of the index for swapping the wrong node numbering with the correct node numbering
-    for values in index_dict.values():
-        for i in range(len(wrongNumberNodes)):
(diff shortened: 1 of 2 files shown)


keiyamamo:
Right now, it is very slow to run the script due to double for loop but I thought it would be better to share it in the repository. The scripts will most likely be updated in the near future.

dbruneau-mie:
Thanks Kei, I will give this a try if I ever have this issue. It has never happened to me, probably an issue specific to SAGA

PR #17, opened 2023-08-04
Refactor and add vampy based meshing script

johannesring:
The new meshing script needs more work, so keeping the old meshing script for now.


keiyamamo (approved):

keiyamamo:
Great! I approved this now and we can modify details later. I will move simulations and post-processing folder after my PR for simulations are merged.

PR #21, opened 2023-08-11
Meshing improvements

johannesring:
Various meshing improvements:
- Added `--meshing-parameters` to control distancetospheres scaling function when using `--meshing-method distancetospheres`.
- Fixed constant solid thickness.
- The mesh is now also saved as a `.h5` file (converted from the XML mesh). This uses DOLFIN.


src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -372,40 +381,45 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
...
-            mesh, remeshed_surface = generate_mesh(distance_to_sphere, add_boundary_layer, meshing_method,
-                                                   solid_thickness, solid_thickness_parameters)
+            mesh, remeshed_surface = generate_mesh(distance_to_sphere, solid_thickness, solid_thickness_parameters)
 
         write_mesh(compress_mesh, file_name_surface_name, file_name_vtu_mesh, file_name_xml_mesh,
                    mesh, remeshed_surface)
 
+        # Add .gz to XML mesh file if compressed
+        if compress_mesh:
+            file_name_xml_mesh = file_name_xml_mesh + ".gz"
```
jorgensd, line 416: I think we should use `pathlib` for filenames in general, then one can use `path.with_suffix(".gz")` to change the file type.

src/fsipy/automatedPreprocessing/preprocessing_common.py
```diff
@@ -160,3 +143,47 @@ def generate_mesh(surface, add_boundary_layer, meshing_method, solid_thickness,
...
+    # Rescale the mesh coordinates
+    x = mesh.coordinates()
+    x[:, :] *= scaling_factor
+    mesh.bounding_box_tree().build(mesh)
+
+    # Convert subdomains to mesh function
+    boundaries = MeshFunction("size_t", mesh, 2, mesh.domains())
+    boundaries.set_values(boundaries.array() + 1)  # FIXME: Explain why this is necessary
+
+    boundary_file = File(str(Path(file_name_xml_mesh).with_suffix('').with_suffix('')) + '_boundaries.pvd')
```
jorgensd, line 176: This is to strip away the `.xml.gz` and add `_boundaries` at the end?
I guess this could be done with:
```python
base, first_dot, rest = file_name_xml_mesh.name.partition('.')
filename = file_name_xml_mesh.with_name(base+"_boundaries.pvd")
```

jorgensd (approved):
Only minor comments regarding pathlib. Looks good !

src/fsipy/automatedPreprocessing/vmtkmeshgeneratorfsi.py
```diff
@@ -374,6 +373,9 @@ def Execute(self):
             tetgen.RegionAttrib = 0
             tetgen.Execute()
 
+            if tetgen.Mesh.GetNumberOfCells() == 0 and surfaceToMesh.Mesh.GetNumberOfCells() > 0:
+                self.PrintLog('An error occurred during tetrahedralization. Will only output surface mesh and boundary layer.')
+
             # ADDING CELL IDs
```
keiyamamo, line 379: @johannesring  Do you think this part can be removed?

keiyamamo (approved):
Looks very good! I put some my thoughts, but please feel free to merge if you want to work on them later. I will test new meshing script soon.

src/fsipy/automatedPreprocessing/automated_preprocessing.py
johannesring: Yes, that would be good.

src/fsipy/automatedPreprocessing/vmtkmeshgeneratorfsi.py
keiyamamo: I meant the part of #ADDING CELL IDS

src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -285,10 +281,18 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
             print("--- Adding flow extensions\n")
             # Add extension normal on boundary for atrium models
             extension = "centerlinedirection" if is_atrium else "boundarynormal"
-            surface_extended = add_flow_extension(surface, centerlines, include_outlet=False,
+            if is_atrium:
```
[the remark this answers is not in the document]
johannesring: We can think about if we want to remove this.

PR #22, opened 2023-08-15
Major update of stenosis problem file 

keiyamamo:
first major update of offset_stenosis problem file. 

- removed `Womersley.py` and use `from vampy.simulation.Womersley import` instead.
- cleaned up the code using `flake8`
- added `.flake8` for configuration


johannesring:
Looks good! I have a few comments:
* Can you add the geometry file (`.vtp` or `.stl`) such that we can generate a mesh and run the problem? I guess a directory `models` in the base directory is a good place to store such files (same as in VaMPY)?
* With the VaMPy based meshing script you will get a `offset_stenosis_info.json` file, where information about inlet and outlet id's are stored. See for example `Artery.py` or `Atrium.py` in VaMPy for how this can be used in the problem file. I am not sure though if the script will select the correct id's for the inlet and the outlet in the stenosis case.

src/fsipy/simulations/offset_stenosis.py
```diff
@@ -112,96 +98,80 @@ def get_mesh_domain_and_boundaries(mesh_file,fsi_region, dx_f_id, fsi_id, rigid_
...
+        self.p_0 = 0.0  # Initial pressure
+        self.P = self.p_0  # Apply initial pressure to inner pressure variable
         super().__init__(**kwargs)
 
     def update(self, t):
         self.t = t
-        # apply a sigmoid ramp to the pressure 
+        # apply a sigmoid ramp to the pressure
         if self.t < self.t_ramp:
-            ramp_factor = (-1/2)*np.cos(3.14159*self.t/self.t_ramp) + 1/2
```
jorgensd, line 158: Should we just use `np.pi` here?

src/fsipy/simulations/offset_stenosis.py
```diff
@@ -218,22 +188,21 @@ def create_bcs(t, v_, DVP, mesh, boundaries, domains, mu_f,
     # Assemble boundary conditions
     bcs = u_inlet + [d_inlet, u_inlet_s, d_inlet_s, d_rigid]
 
-    # Define the pressure condition (apply to inner surface, numerical instability results from applying to outlet)
-    #dso = ds(outlet_id1, domain=mesh, subdomain_data=boundaries) # Outlet surface # Maybe try applying to all outlets???
-
-
-
-    # Load fourier coefficients and scale by flow rate
+    # Load fourier coefficients for the pressure and scale by flow rate
```
jorgensd, line 191:
```suggestion
    # Load Fourier coefficients for the pressure and scale by flow rate
```

src/fsipy/simulations/offset_stenosis.py
```diff
@@ -218,22 +188,21 @@ def create_bcs(t, v_, DVP, mesh, boundaries, domains, mu_f,
...
     An_P, Bn_P = np.loadtxt(os.path.join(os.path.dirname(os.path.abspath(__file__)), P_FC_File)).T
 
+    # Apply pulsatile pressure at the fsi interface by modifying the variational form
+    n = FacetNormal(mesh)
     dSS = Measure("dS", domain=mesh, subdomain_data=boundaries)
     p_out_bc_val = InnerP(t=0.0, t_ramp=0.2, An=An_P, Bn=Bn_P, period=T_Cycle, P_mean=P_mean, degree=p_deg)
-    F_solid_linear += p_out_bc_val * inner(n('+'), psi('+'))*dSS(fsi_id)  # defined on the reference domain
+    F_solid_linear += p_out_bc_val * inner(n('+'), psi('+')) * dSS(fsi_id)
```
jorgensd, line 198: I'm not sure this is safe. As I mentioned in https://github.com/KVSlab/turtleFSI/issues/40 one-sided interior facet integrals needs a volume subdomain to be correctly oriented.

johannesring:
I think that importing `Womersley` from VaMPy is a bit heavy, since it requires morphMan and vmtk to be installed. On a cluster I would like to avoid to install these just to run a turtleFSI simulation. Maybe we can change VaMPy such that these libraries are only imported when importing the pre-processing scripts? Currently, these libraries are imported when doing just `import vampy`, however, they are only used for pre-processing.

keiyamamo:
@johannesring  I completely agree with you. To me, the best is to modify VaMPy so that it imports morphMan and vmtk only when running pre-processing, Oasis only when running simulations. Right now, everything has to be installed and imported even when we want to use parts of the VaMPy.

src/fsipy/simulations/offset_stenosis.py
keiyamamo: I see. We do have subdomain tags. Fluid domain is always marked as 1 and the solid domain is marked as 2. When constructing `F`, we do not have whole domain integral like `*dx` but we do have integral over each domain https://github.com/KVSlab/turtleFSI/blob/34925bed03008e463ce6ddc82b875287fc866177/turtleFSI/modules/fluid.py#L32 and https://github.com/KVSlab/turtleFSI/blob/34925bed03008e463ce6ddc82b875287fc866177/turtleFSI/modules/solid.py#L40. Is that enough to ensure the consistent orientation?

src/fsipy/simulations/offset_stenosis.py
keiyamamo: Agreed and resolved!

src/fsipy/simulations/offset_stenosis.py
jorgensd: As long as `dx` is a ufl Measure with a subdomain_marker with this orientation in it, any integral should be fine.

keiyamamo: Great! Then, I think we are safe here. We have `dx_f[fluid_region] = dx(dx_f_id[fluid_region], subdomain_data=domains)` where `domains` contain subdomain tags for the fluid and solid (1 and 2).

jorgensd: Add that and multiply by a 0 constant, and then things should be consistently ordered only using parameters from the cells with marker 2. (ie normals point out of material 2)

keiyamamo: We do have other terms with `dx` inside `F_solid_linear` as you can see https://github.com/KVSlab/turtleFSI/blob/34925bed03008e463ce6ddc82b875287fc866177/turtleFSI/modules/solid.py#L39-L41 (for example, `(rho_s/k * inner(v_["n"] - v_["n-1"], psi)*dx_s[solid_region]`) 

Shouldn’t that be enough to make sure the consistent orientation? When I add any term with `Constant(0)*dx`, then there is arity mismatch error. 

```
raise ArityMismatch("Adding expressions with non-matching form arguments {0} vs {1}.".format(_afmt(a), _afmt(b)))
ufl.algorithms.check_arities.ArityMismatch: Adding expressions with non-matching form arguments ('v_0',) vs ().
```

johannesring:
@keiyamamo - I think we can work on that later. Just keep in mind that it might not always give the correct ID's.

keiyamamo:
@johannesring - Sounds good! By the way, I used new meshing for stenosis and noticed that domain ID was not correct in `.h5` and `.pvd` file. As you can see in the attached image, all the domain (fluid, solid) was marked as 0. 

I think the problem happens iniside [write_mesh](https://github.com/KVSlab/VaMPy/blob/91605ba23fe9aa5189f9b5a5525d1bef7a46a28b/src/vampy/automatedPreprocessing/preprocessing_common.py#L718C10-L718C10) function in VaMPy when converting from `.vtu` to `.xml`. The output of `domains.array()` before https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/c4d9f88f31c631edc9c9cc5cfaf6b443c61e11d7/src/fsipy/automatedPreprocessing/preprocessing_common.py#L183
is 

```
In [1]: domains.array()
Out[1]: 
array([18446744073709551615, 18446744073709551615, 18446744073709551615,
       ..., 18446744073709551615, 18446744073709551615,
       18446744073709551615], dtype=uint64)
```
and 18446744073709551615 is the maximum value of `size_t`. So my guess is that the domain ID was not correctly passed from `.vtu` to `.xml`. I’m investigating how to fix this and will let you know if I find a fix. 
 Screenshot 2023-08-17 at 12 03 36

keiyamamo:
I figured out changing this line
https://github.com/KVSlab/VaMPy/blob/91605ba23fe9aa5189f9b5a5525d1bef7a46a28b/src/vampy/automatedPreprocessing/preprocessing_common.py#L742
to 
```
writer.SetStoreCellMarkers(1) 
```
should fix the problem. 
stenosis
Will make a PR in VaMPy.

dbruneau-mie (commented):
It is looking good so far. Few comments also mentioned in our meeting:
1. Defining deformable region in terms of x and y
2. Possible mesh changes: shorten inlet, coarsen and use only one boundary layer, try to get running without increasing viscosity at outlet if possible.

keiyamamo:
@dbruneau-mie  Thank you for the helpful feedback! I’ll make sure to incorporate them.

johannesring (approved):
I will merge this now and Kei will work on making the simulation turbulent in a new branch.

PR #27, opened 2023-08-18
add number of sublayer as command line arguments

keiyamamo:
This PR will allow user to choose number of sublayers in the fluid and solid domain from command line arguments. 
I modified parts of the scripts based on flake8.


johannesring (approved):
Looks good.

PR #28, opened 2023-08-21
Add missing docs and add type hints to functions

johannesring:
This PR includes:
- Added missing docs
- Added type hints to functions
- Fixing minor problem with xml mesh to hdf5 conversion


keiyamamo (approved):
Looks great!

PR #31, opened 2023-08-22
Add tests and CI

johannesring:
- Add tests for mesh generation
- Add github workflow (copied from VaMPy)


keiyamamo (approved):
Looks great! I will wait a bit and see if @jorgensd have any comments.

PR #33, opened 2023-08-28
Improve stenosis

keiyamamo:
This PR adds 

1. Improvement of `offset_stenosis.py`. We use x-coordinate to determine the FSI region instead of sphere. 
2. a folder `data/models` and a stl file for the offset stenosis problem. The reason for creating a folder `data` as a parent folder of `models` is that we might want to create other folder for storing data necessary to run simulations (right now we have data for boundary conditions in `src/fsipy/simulations` but should be moved later?)

Please let me know if you have suggestions! For now, the boundary ID in the mesh and ID in `.json` file do not always match, so the boundary IDs are hard-coded. It would be nice to improve this so that it is automated to create boundary conditions.


johannesring:
Looks good. Can you show me the command you use to generate the mesh?

src/fsipy/simulations/offset_stenosis.py
```diff
@@ -64,20 +64,20 @@ def set_problem_parameters(default_variables, **namespace):
...
         dx_s_id=2,  # ID of marker in the solid domain
         # FSI parameters
-        fsi_region=[0.003, 0, 0, 0.01],  # x, y, and z coordinate of FSI region center, and radius of FSI region sphere
+        fsi_x_range=[0.0, 0.016],  # range of x coordinates for fsi region
         # Simulation parameters
         folder="offset_stenosis_results",  # Folder name generated for the simulation
         mesh_path="mesh/file_stenosis.h5",
-        FC_file="FC_MCA_10",  # File name containing the fourier coefficients for the flow waveform
-        P_FC_File="FC_Pressure",  # File name containing the fourier coefficients for the pressure waveform
+        FC_file="FC_MCA_10",  # File name containing the Fourier coefficients for the flow waveform
```
johannesring, line 71: There is no file `FC_MCA_10`. Should this be `FC_file="MCA_10"` or should `MCA_10` be renamed to `FC_MCA_10`?

keiyamamo:
> Looks good. Can you show me the command you use to generate the mesh?

HI @johannesring  The command I used is given [here](https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/268825f1639faea20bb4bf74c9a64965a42d01aa/data/models/offset_stenosis/README.md) :)

src/fsipy/simulations/offset_stenosis.py
keiyamamo: `FC_MCA_10` is the Fourier coefficient version of the `MCA_10`. Not sure if we need to keep both of them, but I think we can discuss this later with others.

keiyamamo:
stenosis

@dbruneau-mie  Could you please review this change? Here is also an animation of the flow field.

keiyamamo:
Hi @dbruneau-mie 
It would be nice if you could have a look on this.
Best,
Kei

src/fsipy/simulations/offset_stenosis.py
```diff
@@ -52,7 +52,7 @@ def set_problem_parameters(default_variables, **namespace):
         P_mean=11200,
         T_Cycle=0.951,  # Used to define length of flow waveform
         rho_f=[1.000E3, 1.000E3],  # Fluid density [kg/m3]
-        mu_f=[3.5E-3, 7.0E-2],  # Fluid dynamic viscosity [Pa.s]
+        mu_f=[1.5E-3, 1.0E-2],  # Fluid dynamic viscosity [Pa.s]
```
dbruneau-mie, line 55: Why the lower blood viscosity here? To help it go turbulent?

dbruneau-mie (commented):
This looks good to me, but it would be nice to have a link or some guide on how to access the mesh/surfaces.

keiyamamo:
@dbruneau-mie I'll add the instructions about the mesh! And yes viscosity is lowered to generate turbulence.

keiyamamo:
@dbruneau-mie @johannesring @jorgensd @KristianValen-Sendstad 

As discussed in the previous meeting, we will move surface and mesh data to `Zenodo`. 
At the same time, I added a `README` file inside `simulations` folder which will explains **where** to get the surface data and **how** to reproduce the mesh. `Zenodo` requires project name, so I will wait until we land on the name of the repository but will add necessary data as soon as we decide the name.

PR #34, opened 2023-08-29
Add VTU to XDMF mesh conversion using meshio

johannesring:
In this pull request a new function `convert_vtu_mesh_to_xdmf` has been added, enabling the conversion of VTU mesh data to the XDMF format using the `meshio` library. Currently, the mesh is converted from VTU to XML and from XML to HDF5. With the change in this PR, the mesh is converted directly from VTU to XDMF. The resulting XDMF mesh can be used in the same way as in [Stenosis_2D.py](https://github.com/KVSlab/turtleFSI/blob/34925bed03008e463ce6ddc82b875287fc866177/turtleFSI/problems/Stenosis_2D.py#L69-L91) in turtleFSI. One thing that is good about using XDMF is that the mesh can easily be viewed in ParaView. Do we want to keep using HDF5 or should we switch to XDMF? Please let me know your thoughts.

### DIFF
--- a/src/fsipy/automatedPreprocessing/preprocessing_common.py
+++ b/src/fsipy/automatedPreprocessing/preprocessing_common.py
@@ -4,6 +4,7 @@ from morphman import vmtkscripts, write_polydata
 from vtk import vtkPolyData
 from dolfin import Mesh, MeshFunction, File, HDF5File
 from pathlib import Path
+import meshio
 
 
 # Global array names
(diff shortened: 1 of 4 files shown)


src/fsipy/automatedPreprocessing/preprocessing_common.py
```diff
@@ -203,3 +204,47 @@ def convert_xml_mesh_to_hdf5(file_name_xml_mesh: str, scaling_factor: float = 0.
     hdf.write(mesh, "/mesh")
     hdf.write(boundaries, "/boundaries")
     hdf.write(domains, "/domains")
+
+
+def convert_vtu_mesh_to_xdmf(file_name_vtu_mesh: str, file_name_xdmf_mesh: str) -> None:
```
jorgensd, line 209: Should we add a note that this is only meant to be executed in serial?
As `meshio` has no concept of MPI it would read the mesh on all processes, which could cause some funky race conditions in parallel.

johannesring:
> They look nice :) I would say that we can keep the functionality to convert to HDF5 for now since a lot of post-processing depends on the HDF5 format.

OK, do you think HDF5 should be the default format for the generated mesh?

johannesring:
I added an option `--mesh-format` to `fsipy-mesh` such that one can choose whether to save the generated mesh as XML, HDF5 or XDMF. The default is HDF5.

keiyamamo (approved):

keiyamamo:
> > They look nice :) I would say that we can keep the functionality to convert to HDF5 for now since a lot of post-processing depends on the HDF5 format.
> 
> OK, do you think HDF5 should be the default format for the generated mesh?

For now, I think we can keep HDF5 as the default, but eventually it would be nice to move to the XDMF. I’m not sure if the post-processing would work in the same way with XDMF at the moment.

PR #35, opened 2023-08-30
Add --flow-rate-factor option

johannesring:
This change adds a `compute_flow_rate` function and a `--flow-rate-factor` option to `fsipy-mesh`. The default value is `0.31`. The function is copied more or less from VaMPy and maybe this change should be added in VaMPy instead.

### DIFF
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -14,13 +14,13 @@ from morphman import get_uncapped_surface, write_polydata, get_parameters, vtk_c
 from vampy.automatedPreprocessing import ToolRepairSTL
 from vampy.automatedPreprocessing.preprocessing_common import read_polydata, get_centers_for_meshing, \
     dist_sphere_diam, dist_sphere_curvature, dist_sphere_constant, get_regions_to_refine, add_flow_extension, \
-    write_mesh, mesh_alternative, find_boundaries, compute_flow_rate, setup_model_network, \
+    write_mesh, mesh_alternative, find_boundaries, setup_model_network, \
     radiusArrayName, scale_surface, get_furtest_surface_point, check_if_closed_surface
 from vampy.automatedPreprocessing.simulate import run_simulation
 from vampy.automatedPreprocessing.visualize import visualize_model
 
 from fsipy.automatedPreprocessing.preprocessing_common import generate_mesh, distance_to_spheres_solid_thickness, \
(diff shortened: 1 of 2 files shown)


keiyamamo:
@johannesring  - Looks nice! Out of curiosity, where does `0.31` come from ? 
@hkjeldsberg  - Is this something that might be useful in `VaMPy`? If so, I think it should go to VaMPy instead.

hkjeldsberg:
@keiyamamo this could definitely be useful in VaMPy too. Regarding the value `0.31` I can see that it is currently set to `0.27` in Vampy, but the source of either of these values are vague. I think one or both of them stem from this paper: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2943236/pdf/nihms232282.pdf – See Table 1, VFR_avg for ICA. I could be wrong though. KVS´paper also uses this value (245 ml/min): https://link.springer.com/article/10.1007/s10439-015-1288-5
So I think this value (245) somehow translates into one of the above.

johannesring:
@keiyamamo - `0.31` is what we have been using in our simulations. @KristianValen-Sendstad can elaborate.
@hkjeldsberg - Ok, I will make a pull request in VaMPy.

keiyamamo:
@johannesring  @hkjeldsberg  Thank you for the explanation!

KristianValen-Sendstad:
Yes, Kei. You are absolutely right. After a rather tedious process of
finding the "correct" (well, least wrong :) flow rates, we concluded that
.31m/s should be the correct velocity to impose on all ICAs in the Aneurisk
Database. Later though, we were a little more cautious since we were
assessing "turbulence" in the ICA and adjusted the value down to .27m/s.
These numbers are so close that it doesn't really matter what we use, but I
think .31m/s is more correct.

KVS

keiyamamo:
@KristianValen-Sendstad  Thank you for the detailed explanation!

johannesring:
I have removed the `compute_flow_rate` function and reverted back to use the one in VaMPY instead. This requires the latest VaMPY from the master branch to work. The `--flow-rate-factor` is still needed in our code and the default flow rate factor is set to `0.31`,

johannesring:
I see that the test fails because it uses VaMPY from before pull request [#120](https://github.com/KVSlab/VaMPy/pull/120) was merged. It looks like this is because the conda environment is cached and I will try to figure out how to clear the cache.

johannesring:
I restarted the failing jobs and now they are all green (I didn't have to clear the cache), so I am merging this now.

PR #38, opened 2023-09-01
Update post-process: 1st step - worked on creating refined mesh with the correct ID markers and ordered nodal coordinate

keiyamamo:
This is work-in-progress and just an example of how we can modify post-process scripts. 
We can look through this together during the meeting. @dbruneau-mie

### DIFF
--- a/src/fsipy/automatedPostprocessing/postprocessing_mesh/create_refined_mesh.py
+++ b/src/fsipy/automatedPostprocessing/postprocessing_mesh/create_refined_mesh.py
@@ -1,28 +1,29 @@
 import numpy as np
 import h5py
-import shutil
 from pathlib import Path
 
 import postprocessing_mesh_common
 
-from dolfin import *
+from dolfin import MPI, Mesh, MeshFunction, HDF5File, refine, adapt, parameters
 
 # This is required to get the boundary refinement to work
 parameters["refinement_algorithm"] = "plaza_with_parent_facets"
 
 """
-This script creates a refined mesh with domain markers from a specified mesh. However, the node numbering may not be the 
+Copyright (c) 2023 David Bruneau
+Modified by Kei Yamamoto 2023
+
+This script creates a refined mesh with domain markers from a specified mesh. However, the node numbering may not be the
 same as the output file with save_deg = 2, so the node numbering is corrected to match the output file
-the output files. Currently, it only runs in serial (not parallel) due to the "adapt" function used in fenics. 
+the output files. Currently, it only runs in serial (not parallel) due to the "adapt" function used in fenics.
 This mesh is later used in  the "postprocessing_h5" and "postprocessing_fenics" scripts.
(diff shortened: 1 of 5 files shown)


keiyamamo:
There are many other things that I need to work on for the post-processing, but this PR itself is ready for review.
Here is a summary of the PR. 

1. I put license identifier at the top of the each file following the syntax of [dolfinx](https://github.com/FEniCS/dolfinx/blob/270438d0bfd165aade0875502c61ea26748083d8/python/dolfinx/__init__.py#L5) Right now, it is not decided how we put license in each files so I thought it would be nice to bring this up. I can make changes to the existing files if we agree on how to put license. 

2. I also put Copyright as you can see in `create_refined_mesh.py`. Copyright is also something we didn’t fully decide what to do, but I put @dbruneau-mie ’s name on this file since it is completely clear that it’s his original work. I don’t know if I should put Copyright in `__init__.py` since there’s very little code in it. If there’s any insight, it would be much appreciated!

3. I modified the original version of `create_refined_mesh.py`. The main purpose of this script is still the same. My work was to be more explicit about what the script is executing and also inserted some checks to reduce errors. Generally, I tried to follow the syntax from @johannesring ’s meshing script. 

This does not have to be merged right now, but would be nice to discuss some details with @dbruneau-mie in the next meeting, and hope to have any comments for improvement from @johannesring  and @jorgensd .

src/fsipy/automatedPostprocessing/postprocessing_mesh/create_refined_mesh.py
```diff
@@ -0,0 +1,181 @@
+# Copyright (c) 2023 David Bruneau
+# Modified by Kei Yamamoto 2023
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+"""
+This script creates a refined mesh with domain markers from a specified mesh. However, the node numbering may not be the
+same as the output file with save_deg = 2, so the node numbering is corrected to match the output file
```
johannesring, line 7: Remove "output file" at the end of this line.

src/fsipy/automatedPostprocessing/postprocessing_mesh/create_refined_mesh.py
```diff
@@ -0,0 +1,181 @@
...
+        indexMap = np.append(orederedWrongNodes, orederedCorrectNodes, axis=1)
+        orderedIndexMap = indexMap[indexMap[:, 0].argsort()]
+        orderedIndexMap = orderedIndexMap[:, [0, 4]]
+
+        # wrongNumberTopology is the topology from the "wrong" numbered mesh.
+        # We will modify this array to change it to the correct node numbering scheme
+        wrongNumberTopology = wrongNumberMesh['mesh/topology'][:]
+
+        # This loop replaces the node numbers in the topology array one by one
+        print('--- Correcting node numbering of the tpology array in the refined mesh \n')
```
johannesring, line 118: Typo her (_tpology_).

johannesring (approved):
Looks good!

keiyamamo:
@johannesring  Thank you for the review!

PR #43, opened 2023-09-05
Add common simulation functions

johannesring:
In this pull request I have added a set of functions in a new file called `simulation_common.py`. These functions provides various functionalities for mesh and data handling, as well as some flow analysis. Here's a short summary of each function:

* `load_mesh_and_data`: Loads mesh, boundary data, and domain data from an HDF5 file.

* `print_mesh_summary`: Prints geometric information about a volumetric mesh.

* `load_mesh_info`: Loads and processes mesh information from a JSON file, returning various parameters like inlet IDs, outlet IDs, wall ID, mean flow rate, area ratios, and inlet areas. The information in the JSON file is currently not always correct, so this function is not so useful yet.

* `load_probe_points`: Loads probe points from a file based on the mesh file's path.

* `print_probe_points`: Prints velocity and pressure at specified probe points.

* `peval`: Allows parallel synchronized evaluation of a function. This is used for probes.

* `calculate_and_print_flow_properties`: Calculates and prints flow properties such as flow rate, velocity, CLF numbers and Reynolds numbers.

### DIFF
--- a/src/fsipy/simulations/simulation_common.py
+++ b/src/fsipy/simulations/simulation_common.py
@@ -87,26 +87,21 @@ def print_mesh_summary(mesh: Mesh) -> None:
 
         # Print various mesh statistics
         print("=== Mesh Information Summary ===")
-        print("X range: {} to {} (delta: {:.4f})"
-              .format(combined_info['x_min'], combined_info['x_max'],
-                      combined_info['x_max'] - combined_info['x_min']))
-        print("Y range: {} to {} (delta: {:.4f})"
-              .format(combined_info['y_min'], combined_info['y_max'],
-                      combined_info['y_max'] - combined_info['y_min']))
-        print("Z range: {} to {} (delta: {:.4f})"
-              .format(combined_info['z_min'], combined_info['z_max'],
-                      combined_info['z_max'] - combined_info['z_min']))
-        print("Number of cells: {}".format(combined_info['num_cells']))
-        print("Number of cells per processor: {}"
-              .format(int(np.mean(num_cells_per_processor))))
-        print("Number of edges: {}".format(combined_info['num_edges']))
-        print("Number of faces: {}".format(combined_info['num_faces']))
-        print("Number of facets: {}".format(combined_info['num_facets']))
-        print("Number of vertices: {}".format(combined_info['num_vertices']))
-        print("Volume: {}".format(volume))
-        print("Number of cells per volume: {}"
-              .format(combined_info['num_cells'] / volume))
-        print()
+        print(f"X range: {combined_info['x_min']} to {combined_info['x_max']} "
+              f"(delta: {combined_info['x_max'] - combined_info['x_min']:.4f})")
+        print(f"Y range: {combined_info['y_min']} to {combined_info['y_max']} "
+              f"(delta: {combined_info['y_max'] - combined_info['y_min']:.4f})")
+        print(f"Z range: {combined_info['z_min']} to {combined_info['z_max']} "
+              f"(delta: {combined_info['z_max'] - combined_info['z_min']:.4f})")
+        print(f"Number of cells: {combined_info['num_cells']}")
+        print(f"Number of cells per processor: {int(np.mean(num_cells_per_processor))}")
+        print(f"Number of edges: {combined_info['num_edges']}")
+        print(f"Number of faces: {combined_info['num_faces']}")
+        print(f"Number of facets: {combined_info['num_facets']}")
+        print(f"Number of vertices: {combined_info['num_vertices']}")
+        print(f"Volume: {volume}")
+        print(f"Number of cells per volume: {combined_info['num_cells'] / volume}\n")
+
 
 
 def load_mesh_info(mesh_path: str) -> Tuple[List[int], List[int], int, float, List[float], List[float]]:
--- a/src/fsipy/simulations/simulation_common.py
+++ b/src/fsipy/simulations/simulation_common.py
@@ -8,12 +8,12 @@ from dolfin import Mesh, assemble, Constant, MPI, HDF5File, Measure, inner, Mesh
     Function, sqrt, Expression, TrialFunction, TestFunction, LocalSolver, dx
 
 
-def load_mesh_and_data(mesh_path: str) -> Tuple[Mesh, MeshFunction, MeshFunction]:
+def load_mesh_and_data(mesh_path: Union[str, Path]) -> Tuple[Mesh, MeshFunction, MeshFunction]:
     """
     Load mesh, boundary data, and domain data from an HDF5 file.
 
     Args:
-        mesh_path (str): Path to the HDF5 file containing mesh and data.
+        mesh_path (str or Path): Path to the HDF5 file containing mesh and data.
 
     Returns:
         Tuple[Mesh, MeshFunction, MeshFunction]:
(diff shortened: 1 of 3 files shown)


src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
+import json
+from typing import List, Union, Tuple
+from pathlib import Path
+
+import numpy as np
+from mpi4py import MPI as mpi
```
keiyamamo, line 6: Should we have mpi4py as a dependency?

src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
...
+        n (dolfin.Expression): FacetNormal expression.
+        dsi (dolfin.Measure): Measure for inlet boundary.
+
+    Returns:
+        None
+    """
+    # Calculate the DG vector of velocity magnitudes
+    DG = FunctionSpace(mesh, "DG", 0)
+    V_vector = project(sqrt(inner(v, v)), DG).vector().get_local()
+    h = mesh.hmin()
```
keiyamamo, line 236: This `calculate_and_print_flow_properties` is definitely useful, but I think we should avoid using `project` inside the function that might be used many times during the time loop. This is because `project` is a very costly function   that can be easily optimized. For reference, @jorgensd ’s [tutorial](https://simula-sscp.github.io/SSCP_2023_lectures/L12%20%28FEniCS%20Intro%29/L03_FEniCS_Darcy.html#advanced-reusable-projection) shows how to make reusable projection.

src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
...
+        print("Number of cells: {}".format(combined_info['num_cells']))
+        print("Number of cells per processor: {}"
+              .format(int(np.mean(num_cells_per_processor))))
+        print("Number of edges: {}".format(combined_info['num_edges']))
+        print("Number of faces: {}".format(combined_info['num_faces']))
+        print("Number of facets: {}".format(combined_info['num_facets']))
+        print("Number of vertices: {}".format(combined_info['num_vertices']))
+        print("Volume: {}".format(volume))
+        print("Number of cells per volume: {}"
+              .format(combined_info['num_cells'] / volume))
```
keiyamamo, line 108: Should we maybe use `f-strings` for consistency here?

keiyamamo (approved):
Looks very nice! I’ve been meaning to bring up the usage of `project` for computing CFL number but didn’t have a chance to do it. Yet, I noticed it was extremely slow when I used it for post-processing, so I would really like to avoid using it especially with turtle. Please let me know if you want me to work on this.

src/fsipy/simulations/simulation_common.py
jorgensd: `mpi4py` is a dependency of Dolfin. so no need to add it explictly

src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
...
+from typing import List, Union, Tuple
+from pathlib import Path
+
+import numpy as np
+from mpi4py import MPI as mpi
+from dolfin import Mesh, assemble, Constant, MPI, HDF5File, Measure, project, inner, MeshFunction, FunctionSpace, \
+    Function, sqrt, Expression
+
+
+def load_mesh_and_data(mesh_path: str) -> tuple:
```
jorgensd, line 11: Do we want to do `tuple[dolfin.Mesh, dolfin.MeshFunction, dolfin.MeshFunction]`, or 
```python
from typing import Tuple
Tuple[...]
```
for python < 3.9?

src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
...
+        uu = peval(v.sub(0), point)
+        vv = peval(v.sub(1), point)
+        ww = peval(v.sub(2), point)
+        pp = peval(p, point)
+
+        if MPI.rank(MPI.comm_world) == 0:
+            print(f"Probe Point {i}: Velocity: ({uu}, {vv}, {ww}) | Pressure: {pp}")
+
+
+def peval(f: Function, x: Union[float, np.ndarray]) -> np.ndarray:
```
jorgensd, line 193: Are we here assuming that the probe points `x` are the same on all processes? what happens if they are not? Then the allreduce doesn't work I guess?

src/fsipy/simulations/simulation_common.py
jorgensd: Especially if we  are projecting into `DG`-0 we should use the local projection, as no global inversion of any matrices are needed.

jorgensd (changes requested):
Some minor comments. In general a very nice and well written PR.

In general I wonder if we should switch print statements with a logger, as it would make it easier to turn on and off such statements.

johannesring:
Thank you both for your comments! I have addressed some of them now and will look at the other comments later.

src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
...
+        "z_max": local_z_coords.max(),
+        "num_cells": mesh.num_cells(),
+        "num_edges": mesh.num_edges(),
+        "num_faces": mesh.num_faces(),
+        "num_facets": mesh.num_facets(),
+        "num_vertices": mesh.num_vertices()
+    }
+
+    # Gather local information from all processors to processor 0
+    comm = MPI.comm_world
```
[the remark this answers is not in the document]
johannesring: I have switched to use the mesh mpi_comm instead.

src/fsipy/simulations/simulation_common.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Open the HDF5 file in read-only mode
+    hdf5 = HDF5File(MPI.comm_world, mesh_path, "r")
+
+    # Read mesh data
+    hdf5.read(mesh, "/mesh", False)
+
+    # Create MeshFunction objects for boundaries and domains
+    boundaries = MeshFunction("size_t", mesh, 2)
+    domains = MeshFunction("size_t", mesh, 3)
```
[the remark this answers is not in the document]
johannesring: I think using `mesh.topology().dim()` is good. I have made this change now.

src/fsipy/simulations/simulation_common.py
johannesring: I went for the latter.

src/fsipy/simulations/simulation_common.py
johannesring: Yes, the probe point `x` are the same on all processes.

src/fsipy/simulations/simulation_common.py
johannesring: @keiyamamo and @jorgensd - I have added Jørgen's `local_project` function and use this instead of `project` in `calculate_and_print_flow_properties`. Let me know if this is the correct way to do it.

jorgensd: Looks good to me.

keiyamamo: @jorgensd  Would it be possible to combine reusable and local projection? or Doesn’t it really make difference in case of local projection ?  I can see that `local_project` function still creates `a_proj` every time even though it is constant.

jorgensd: No, i couldnt find a way of doing the Pre-assembly of the lhs with localsolver. This is a limitation of Dolfin. 

One could have an option to either use a global matrix (the reusable projector, that works with any function space), or the local projector (that only works on DG elements.

jorgensd: Yes, But it would require some testing, as I not 100% confident local solve is always the right thing to use. We might be better of using a re-usable (global) projection class if we want things to be guaranteed to do what we expect.

keiyamamo: I see. I just want to make sure that projection is as cheap as possible since this function `calculate_and_print_flow_properties` is supposed to be called many times during the simulation. I might compute the time difference between local and re-usable(global) projection and see which one is faster. For now, I think we can keep what Johannes made :)

johannesring:
I think we have addressed all the comments. Is this ready to be merged? @jorgensd

jorgensd (approved):

PR #47, opened 2023-09-11
add a script for separating domains

keiyamamo:
This PR adds a script named `separate_domain.py`, which is basically a combination of `Create_Fluid_Only_Mesh.py` and `Create_Solid_Only_Mesh.py`. 

I did my best to optimize the code because [fixing the topology ](https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/1e6ea2313c1161fb83e70c77ec2fe264aaeb3023/postprocessing_mesh/common_meshing.py#L68-L69)was very computationally expensive and it is much faster now. (line 69-74 of `separate_domain.py`). However, it still requires to loop through all the node of each domain and takes some time for a large mesh. If anyone has suggestion how to avoid for loop, it would be very much appreciated.

Additionally, this script read `default_variable.pickle` to extract domain ID for fluid and solid and utilize these information to separate the domain. This avoids both command-line argument and hard-coded domain ID. I believe other post-processing script can also follow this style to avoid user inputs as much as possible.


src/fsipy/automatedPostprocessing/postprocessing_mesh/separate_domain.py
```diff
@@ -0,0 +1,143 @@
...
+import h5py
+from pathlib import Path
+
+from fsipy.automatedPostprocessing.postprocessing_mesh import postprocessing_mesh_common
+
+from dolfin import MPI, Mesh, MeshFunction, HDF5File, SubMesh, File
+
+
+def separate_domain(mesh_path: Path, fluid_domain_id: int, solid_domain_id: int, view: bool = False) -> None:
+    """
```
jorgensd, line 16: Could you add a doc-string explaining the purpose of this function?

src/fsipy/automatedPostprocessing/postprocessing_mesh/separate_domain.py
```diff
@@ -0,0 +1,143 @@
...
+        print(f" --- Fixing topology of {domain_name} domain \n")
+        print(" --- This is computationally expensive \n")
+        if not np.all(np.diff(domain_of_interest_ids) == 1):
+            for node_id, domain_id in enumerate(domain_of_interest_ids):
+                if node_id % 1000 == 0:
+                    print(f" --- Fixing topology of {domain_name} domain: {node_id} / {len(domain_of_interest_ids)} \n")
+                mask = domain_of_interest_topology == domain_id
+                domain_of_interest_topology[mask] = node_id
+        else:
+            print(f" --- {domain_name} topology does not need to be fixed \n")
```
jorgensd, line 76: @keiyamamo could you drop by at some point so we can discuss this?

keiyamamo: @jorgensd  Of course!

src/fsipy/automatedPostprocessing/postprocessing_mesh/separate_domain.py
```diff
@@ -0,0 +1,143 @@
...
+        # This is necessary because the node numbering may not be continuous
+        # while there is one to one correspondence between the node IDs and the coordinates
+        print(f" --- Fixing topology of {domain_name} domain \n")
+        print(" --- This is computationally expensive \n")
+        if not np.all(np.diff(domain_of_interest_ids) == 1):
+            for node_id, domain_id in enumerate(domain_of_interest_ids):
+                if node_id % 1000 == 0:
+                    print(f" --- Fixing topology of {domain_name} domain: {node_id} / {len(domain_of_interest_ids)} \n")
+                mask = domain_of_interest_topology == domain_id
+                domain_of_interest_topology[mask] = node_id
```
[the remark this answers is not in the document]
keiyamamo: Amazing! I tested this and the results seem identical, and much faster :) Will incorporate this.

PR #52, opened 2023-09-13
Add simple test cylinder

dbruneau-mie:
Simple test cylinder added to repo.

### DIFF
--- a/src/fsipy/simulations/cylinder.py
+++ b/src/fsipy/simulations/cylinder.py
@@ -38,7 +38,7 @@ def set_problem_parameters(default_variables, **namespace):
             recompute=20,  # Recompute the Jacobian matix within time steps
             recompute_tstep=20,  # Recompute the Jacobian matix over time steps
             # boundary condition parameters
-            mesh_path="artery_coarse_rescaled",
+            mesh_path="mesh/cylinder.h5",
             inlet_id=2,  # inlet id for the fluid
             inlet_outlet_s_id=11,  # inlet and outlet id for solid
             fsi_id=22,  # id for fsi interface
@@ -59,8 +59,7 @@ def set_problem_parameters(default_variables, **namespace):
             # mesh lifting parameters (see turtleFSI for options)
             extrapolation="laplace",  # laplace, elastic, biharmonic, no-extrapolation
             extrapolation_sub_type="constant",  # ["constant","small_constant","volume","volume_change","bc1","bc2"]
-            folder="cylinder_test",  # output folder generated for simulation
-            kill_time=100000,  # in seconds, after this time start dumping checkpoints every timestep
+            folder="cylinder_results",  # output folder generated for simulation
             save_deg=1  # save_deg=1 saves corner nodes only, save_deg=2 saves corner + mid-point nodes for viz
         )
     )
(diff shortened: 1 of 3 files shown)


src/fsipy/simulations/cylinder.py
```diff
@@ -1,3 +1,223 @@
...
+        )
+    )
+
+    return default_variables
+
+
+def get_mesh_domain_and_boundaries(mesh_path, folder, **namespace):
+    print("Obtaining mesh, domains and boundaries...")
+    mesh = Mesh()
+    hdf = HDF5File(mesh.mpi_comm(), "mesh/" + mesh_path + ".h5", "r")
```
keiyamamo, line 74:
```suggestion
    hdf = HDF5File(mesh.mpi_comm(), mesh_path, "r")
```
Instead of assuming that there’s a folder `mesh`, it would be more user-friendly to directly use path to the mesh.

src/fsipy/simulations/cylinder.py
```diff
@@ -1,3 +1,223 @@
...
+    P_final,
+    v_max_final,
+    fsi_id,
+    inlet_id,
+    inlet_outlet_s_id,
+    rigid_id,
+    psi,
+    F_solid_linear,
+    **namespace
+):
```
keiyamamo, line 169: I’m guessing you followed `flake8`, but changing line for each variables is a bit overkill and also makes overall script longer than necessary. You can have a few arguments and move to the next line when it becomes too long. For example, 
https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/268825f1639faea20bb4bf74c9a64965a42d01aa/src/fsipy/simulations/offset_stenosis.py#L158-L161

src/fsipy/simulations/cylinder.py
```diff
@@ -1,3 +1,223 @@
...
+
+    # Parabolic Inlet Velocity Profile
+    u_inflow_exp = VelInPara(
+        t=0.0,
+        t_ramp=0.1,
+        v_max_final=v_max_final,
+        n=normal,
+        dsi=dsi,
+        mesh=mesh,
+        degree=3,
```
keiyamamo, line 194: Same comment as above.

src/fsipy/simulations/cylinder.py
```diff
@@ -1,3 +1,223 @@
...
+    )
+    u_inlet = DirichletBC(DVP.sub(1), u_inflow_exp, boundaries, inlet_id)
+    u_inlet_s = DirichletBC(
+        DVP.sub(1), ((0.0, 0.0, 0.0)), boundaries, inlet_outlet_s_id
+    )
```
keiyamamo, line 199: If you run `flake8` again with new configuration, I think this `u_inlet_s` can be just one line. Could you consider that? Having only `)` is a bit confusing.

src/fsipy/simulations/cylinder.py
```diff
@@ -1,3 +1,223 @@
...
+
+    # Solid Displacement BCs
+    d_inlet = DirichletBC(DVP.sub(0), ((0.0, 0.0, 0.0)), boundaries, inlet_id)
+    d_inlet_s = DirichletBC(
+        DVP.sub(0), ((0.0, 0.0, 0.0)), boundaries, inlet_outlet_s_id
+    )
```
keiyamamo, line 205: Same as above.

keiyamamo (changes requested):

keiyamamo (approved):

keiyamamo:
@dbruneau-mie I talked with @jorgensd about the authorship for each functions and we concluded that listing all the authors' name at the top of script is better than listing author name for each functions. If we include author name as doc-string in each functions, user will be exposed to the author name when using `help(some_function)` and we thought that information might not be the most critical one for the user. 

You do not need to change anything for this pull request, but we will put all the names at the beginning of the script when we  will work on these scripts that you added. Please let me know if you have any thoughts :)

PR #59, opened 2023-09-21
Add test for simulations folder

keiyamamo:
With this PR, I created a test for running problem file with `turtleFSI`. First, I modified `offset_stenosis.py` so that it loads probe points and print out the velocity and pressure during the execution. In the test file, `turtleFSI` runs 5 time step and check if the printed velocity and pressure matches the expected value. Right now, I only added a test for offset_stenosis, but  will add test for cylinder.


keiyamamo:
Hi @jorgensd  

This pytest runs with no problem on my laptop but fails on GitHub. The error message is 

> *** -------------------------------------------------------------------------
  *** Error:   Unable to solve linear system using PETSc Krylov solver.
  *** Reason:  Solution failed to converge in 0 iterations (PETSc reason DIVERGED_PC_FAILED, residual norm ||r|| = 0.000000e+00).
  *** Where:   This error was encountered inside PETScKrylovSolver.cpp.
  *** Process: 0
  *** DOLFIN version: 2019.1.0
  *** Git changeset:  12ef077802cc9fad34cf984ec7af80585b44301b
  *** -------------------------------------------------------------------------
  
  which is the same error message I used to get on Saga (https://turtlefsi2.readthedocs.io/en/latest/known_issues.html). The issue on Saga was resolved by increasing the memory allocation with MUMPS but I’m not sure if that can be used here since GitHub seem to have limited memory. Do you have any suggestion how to make this test run ?

keiyamamo:
Adding a bit more information. I rerun the test with osx-64 on my forked repository and the test passed with no problem. 
https://github.com/keiyamamo/Aneurysm_Workflow_FSI/actions/runs/6265951848/job/17016292263?pr=7
According to the [documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), macOS has more RAM than linux machine, so might explain why the test runs with osx-64 but not linux. I will try to make mesh smaller and see if that helps.

keiyamamo:
closing for now since the test is not passing on github. I will reopen once I fix the problem.

Issue #57, opened 2023-09-18
Remove generated mesh files when running pre-processing test


keiyamamo:
Running `test_pre_processing.py` creates quite many files and they are not automatically removed after the test is finished.
Could it be possible to improve the test so that newly generated files are removed after the test?

johannesring:
Yes, I will fix this when I am back from vacation on Friday.

keiyamamo:
Fixed #61

PR #61, opened 2023-09-23
Refactor pre-processing test paths to use a temporary directory

johannesring:
Refactor test paths to use a temporary directory for pre-processing tests. This change ensures that tests run in isolation without affecting the original data. Fixes issue #57.


keiyamamo (approved):
Looks good! Thank you for the improvement!

PR #62, opened 2023-09-25
Add log file parsing and plotting script

johannesring:
This Python script provides a flexible solution for parsing and visualizing simulation log files created by `turtleFSI`. It enables users to extract data from log files, offers flexible plotting options, and helps in gaining insights from simulation results.

Usage Example:

``` shell
fsipy-log-plotter simulation.log --plot-all --save --output-directory Results --figure-size 12,8
```
Functions:

* `parse_log_file`: Parses a log file and stores data in a structured dictionary with NumPy arrays.
* `parse_dictionary_from_log`: Parses a dictionary-like content from a log file and returns it as a dictionary.
* `plot_variable_vs_time`: Plots a variable against time with optional saving of the figure.
* `plot_variable_comparison`: Plots a comparison of a variable across multiple cycles.
* `plot_multiple_variables_vs_time`: Plots mean, min, and max variables against time on the same figure.
* `plot_multiple_variables_comparison`: Compares mean, min, and max variables across multiple cycles.
* `plot_newton_iteration`: Plots Newton iteration (atol or rtol).
* `plot_probe_points`: Plots velocity magnitude and pressure for probe points against time.
* `plot_probe_points_comparison`: Compares velocity magnitude and pressure for probe points across multiple cycles.
* `save_plot_to_file`: Saves a plot to a file.
* `compute_average_over_cycles`: Computes the average over cycles in a NumPy array.
* `parse_command_line_args`: Parses command-line arguments for the script's configuration.

Note: Detailed usage and options can be found by running the script with the `--help` flag.

### DIFF
--- a/src/fsipy/automatedPostprocessing/log_plotter.py
+++ b/src/fsipy/automatedPostprocessing/log_plotter.py
@@ -425,7 +425,7 @@ def plot_newton_iteration(variable: np.ndarray, variable_name: str, save_to_file
     logging.info(f"--- Creating plot for {variable_name}")
 
     plt.figure(figsize=figure_size)
-    plt.plot(variable, label=variable_name, linestyle='-', color='b')
+    plt.plot(variable, label=variable_name, linestyle='--', marker='o', color='b')
     plt.ylabel(variable_name)
     plt.title(f"{variable_name}")
     plt.grid(True)
--- a/src/fsipy/automatedPostprocessing/log_plotter.py
+++ b/src/fsipy/automatedPostprocessing/log_plotter.py
@@ -15,6 +15,7 @@ fsipy-log-plotter simulation.log --plot-all --save --output-directory Results --
 """
 
 import re
+import json
 import argparse
 import logging
 from pathlib import Path
@@ -195,17 +196,22 @@ def parse_dictionary_from_log(log_file: str) -> dict:
 
                 # Check if the line ends the current dictionary entry
                 if line.endswith('}'):
-                    # Combine the lines and evaluate as a dictionary
+                    # Combine the lines and modify for JSON compatibility
                     entry_str = '\n'.join(entry_lines)
+                    entry_str = entry_str.replace("'", '"').replace("None", "null").replace("True", "true").replace("False", "false")
+
+                    # Handle PosixPath objects by converting them to strings
+                    entry_str = re.sub(r'"(restart_folder)":\s+PosixPath\("([^"]+)"\)', r'"\1": "\2"', entry_str)
+
                     try:
-                        entry_dict = eval(entry_str)
+                        entry_dict = json.loads(entry_str)
                         if isinstance(entry_dict, dict):
                             parsed_dict.update(entry_dict)
                             break  # Exit the loop since there is only one dictionary
                         else:
                             logging.warning(f"WARNING: Entry is not a valid dictionary: {entry_str}")
-                    except SyntaxError:
-                        logging.warning(f"WARNING: SyntaxError while parsing entry: {entry_str}")
+                    except json.JSONDecodeError as e:
+                        logging.warning(f"WARNING: JSONDecodeError while parsing entry: {e}")
 
     return parsed_dict
 
--- a/src/fsipy/automatedPostprocessing/log_plotter.py
+++ b/src/fsipy/automatedPostprocessing/log_plotter.py
@@ -709,7 +709,7 @@ def main() -> None:
 
     # Calculate the number of cycles and the number of time steps per cycle
     num_cycles = int(end_time / cycle_length)
-    time_steps_per_cycle = round(cycle_length / dt) + 1
+    time_steps_per_cycle = round(cycle_length / dt)
 
     # Determine start and end range for data
     start_cycle = args.start_cycle
--- a/src/fsipy/automatedPostprocessing/log_plotter.py
+++ b/src/fsipy/automatedPostprocessing/log_plotter.py
@@ -763,7 +763,14 @@ def main() -> None:
         start = 0
         end = len(cpu_time)
 
-    if args.plot_all or args.plot_cpu_time:
+    def check_and_warn_empty(variable_name, variable_data, condition):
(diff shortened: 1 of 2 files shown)


keiyamamo:
Hi @johannesring 

Looks great! I have tested this function locally and have some thoughts. 

1. When I tried to run the `log_plotter` with restarted log file, I got an error
```
entry_dict = eval(entry_str)
  File "<string>", line 89, in <module>
NameError: name 'PosixPath' is not defined
```
which comes from 
```
'restart_folder': PosixPath('/cluster/work/users/keiya/case16/1/Checkpoint'),
```
in the log file. I guess we need to convert `restart_folder` to `str` type before printing in `turtleFSI` ?

2. I personally prefer plotting with market for newton iterations. With only line, it is a bit hard for me to understand how many iterations we actually have for the newton iteration to converge. Below is an example of plot with marker. With marker, we can count the number of newton iterations visually.
Figure_1

```diff
+ plt.plot(variable, label=variable_name, linestyle='--', marker='o', color='b’)
- plt.plot(variable, label=variable_name, linestyle='-', color='b’)
```
Same for CPU time, but I figured out it might be tricky since we are using `plot_variable_vs_time` for CPU time which accepts other variables.

johannesring:
> 1. When I tried to run the `log_plotter` with restarted log file, I got an error
> 
> ```
> entry_dict = eval(entry_str)
>   File "<string>", line 89, in <module>
> NameError: name 'PosixPath' is not defined
> ```
> 
> which comes from
> 
> ```
> 'restart_folder': PosixPath('/cluster/work/users/keiya/case16/1/Checkpoint'),
> ```
> 
> in the log file. I guess we need to convert `restart_folder` to `str` type before printing in `turtleFSI` ?

I have fixed this problem now. I also changed from using `eval` to `json.loads` when parsing the dictionary. Feel free to also update turtleFSI to convert `restart_folder` to `str`.

> 2. I personally prefer plotting with market for newton iterations. With only line, it is a bit hard for me to understand how many iterations we actually have for the newton iteration to converge. Below is an example of plot with marker. With marker, we can count the number of newton iterations visually.
>    Figure_1
> 
> ```diff
> + plt.plot(variable, label=variable_name, linestyle='--', marker='o', color='b’)
> - plt.plot(variable, label=variable_name, linestyle='-', color='b’)
> ```

I agree with this and I have made the change.

> Same for CPU time, but I figured out it might be tricky since we are using `plot_variable_vs_time` for CPU time which accepts other variables.

We could consider adding `linestyle` and `marker` as arguments to the `plot_variable_vs_time` function. Let me know what you think about that.

johannesring:
> Lastly, I tried newer version with `compute-average` flag and I got an error
> 
> ```
> line 631, in compute_average_over_cycles
> data_3d = data[:total_cycles * time_steps_per_cycle].reshape(total_cycles, time_steps_per_cycle, -1)
> ValueError: cannot reshape array of size 0 into shape (0,5601,newaxis)
> ```
> 
> This is because my log file does not have all the variables. Would it be possible to add `check_and_warn_empty` for `compute-average` as well? Please let me know what you think :)

I have made a fix for this now. Please let me know if it works or not.

src/fsipy/automatedPostprocessing/log_plotter.py
```diff
@@ -0,0 +1,908 @@
...
+    parsed_data = parse_log_file(args.log_file)
+
+    # Extract end time, cycle length and time step size from the parsed dictionary
+    end_time = parsed_dict.get("T", 0.951)
+    cycle_length = parsed_dict.get("T_cycle", 0.951)
+    dt = parsed_dict.get("dt", 0.001)
+
+    # Calculate the number of cycles and the number of time steps per cycle
+    num_cycles = int(end_time / cycle_length)
+    time_steps_per_cycle = round(cycle_length / dt) + 1
```
keiyamamo, line 712: In my case, I had to remove this +1 for averaging to work properly. I had ’T=2.783’ and `dt=0.00016982142` to have 3 cycles with 5600 time steps per cycle. Could you check if +1 is necessary?

src/fsipy/automatedPostprocessing/log_plotter.py
```diff
@@ -0,0 +1,908 @@
...
+        logging.warning("WARNING: Select either --compute-average or --compare-cycles, not both.")
+        args.compare_cycles = False
+
+    # Parse log data
+    parsed_dict = parse_dictionary_from_log(args.log_file)
+    parsed_data = parse_log_file(args.log_file)
+
+    # Extract end time, cycle length and time step size from the parsed dictionary
+    end_time = parsed_dict.get("T", 0.951)
+    cycle_length = parsed_dict.get("T_cycle", 0.951)
```
keiyamamo, line 707: If `T_cycle` is based on offset_stenosis problem, it should be `T_Cycle`? maybe I need to fix the offset_stenosis since using capital letter is a bit weird...

keiyamamo (commented):
Thanks again! I still have few comments but they all look really nice!

src/fsipy/automatedPostprocessing/log_plotter.py
johannesring: Yes, it is based on the offset stenosis problem. In VaMPy, `cardiac_cycle` is used in `Artery.py` and `Atrium.py`. Should we also use that?

src/fsipy/automatedPostprocessing/log_plotter.py
johannesring: You are right, it looks like it is not necessary. I have removed it now.

src/fsipy/automatedPostprocessing/log_plotter.py
keiyamamo: Yes, `cardiac_cycle` seems better!

johannesring: Ok, I will update the script to use that instead.

keiyamamo: Thanks!

keiyamamo (approved):

johannesring:
Thanks for merging @keiyamamo!

PR #65, opened 2023-09-28
Include mesh IDs in parameter file

johannesring:
This change adds new command line arguments to `fsipy-mesh`, specifically `--solid-side-wall-id`, `--interface-fsi-id`, `--interface-outer-id`, `--volume-id-fluid`, and `--volume-id-solid`. It also adds the functionality to write these ID values to the parameters file (`_info.json`). The `generate_mesh` function in `preprocessing_common.py` now also accepts these ID values as arguments. The change also includes modifications to the `load_mesh_info` function in `simulation_common.py` to return a named tuple `MeshInfo` containing all relevant mesh information, including the new ID values. The test functions have also been updated to reflect these changes.

### DIFF
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -28,8 +28,8 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
                        coarsening_factor, inlet_flow_extension_length, outlet_flow_extension_length,
                        number_of_sublayers_fluid, number_of_sublayers_solid, edge_length,
                        region_points, compress_mesh, scale_factor, resampling_step, meshing_parameters,
-                       remove_all, solid_thickness, solid_thickness_parameters, mesh_format,
-                       flow_rate_factor):
+                       remove_all, solid_thickness, solid_thickness_parameters, mesh_format, flow_rate_factor,
+                       solid_side_wall_id, interface_fsi_id, interface_outer_id, volume_id_fluid, volume_id_solid):
     """
     Automatically generate mesh of surface model in .vtu and .xml format, including prescribed
     flow rates at inlet and outlet based on flow network model.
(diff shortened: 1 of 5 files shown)


src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -439,6 +444,14 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
             # Write mesh in VTU format
             write_polydata(remeshed_surface, file_name_surface_name)
             write_polydata(mesh, file_name_vtu_mesh)
+
+        # Write the mesh ID's to parameter file
+        parameters["solid_side_wall_id"] = solid_side_wall_id
+        parameters["interface_fsi_id"] = interface_fsi_id
+        parameters["interface_outer_id"] = interface_outer_id
```
keiyamamo, line 451: Should this be `solid_outer_wall_id`? To me, the word `interface_outer` is non-intuitive.

src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -439,6 +444,14 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
...
+        parameters["volume_id_fluid"] = volume_id_fluid
+        parameters["volume_id_solid"] = volume_id_solid
```
keiyamamo, line 453: If we follow the naming above, it might be better to call them `volume_fluid/solid_id`

keiyamamo (approved):
Thanks for the update! I have a few suggestions for naming.

src/fsipy/automatedPreprocessing/automated_preprocessing.py
johannesring: I agree.

src/fsipy/automatedPreprocessing/automated_preprocessing.py
johannesring: Perhaps even `fluid_volume_id` and `solid_volume_id`? What do you like better?

keiyamamo: `fluid_volume_id` and `solid_volume_id` sound great!

PR #68, opened 2023-10-02
Add TKE visualization for probe plots

johannesring:
This pull request improves the existing Python script for parsing and visualizing data from log files. The key changes include the addition of new functions and options related to Turbulent Kinetic Energy (TKE) calculations for probe points:

* Function `compute_tke`: Computes mean velocity, fluctuating velocity, and TKE for each point within a specified cycle range based on probe point data.
* Function `plot_probe_points_tke`: Plots TKE for each probe point in separate subplots.
* Function `plot_probe_points_tke_comparison`: Plots TKE per cycle for selected probe points in separate figures.
* Added `--plot-probe-points-tke` option to enable plotting of TKE for probe points. This can be used in combination with either `--compare-cycles` or ` --compute-average`.

### DIFF
--- a/src/fsipy/automatedPostprocessing/log_plotter.py
+++ b/src/fsipy/automatedPostprocessing/log_plotter.py
@@ -650,7 +650,7 @@ def compute_tke(probe_points: Dict[int, Dict[str, Any]], time_steps_per_cycle: i
                 end_cycle: Optional[int] = None) -> Dict[int, Tuple[np.ndarray, np.ndarray, np.ndarray]]:
     """
     Compute the mean velocity, fluctuating velocity, and turbulent kinetic energy for each point within the specified
-    cycle range in the probe points data.
+    cycle range in the probe points data using phase-averaging.
 
     Args:
         probe_points (dict): Probe points data as returned by parse_log_file.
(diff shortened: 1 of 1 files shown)


src/fsipy/automatedPostprocessing/log_plotter.py
```diff
@@ -505,6 +508,11 @@ def plot_probe_points(time: np.ndarray, probe_points: Dict[int, Dict[str, np.nda
         ax2.legend([l1, l2], ["Velocity Magnitude", "Pressure"], loc="upper right")
         ax2.tick_params(axis='y', which='major', labelcolor='r')
 
+    # Remove any empty subplots if the number of probe points doesn't fill the entire grid
+    for i in range(num_selected_probe_points, num_rows * num_cols):
```
keiyamamo, line 512: I tried this new functions but got a following error here.
```
--- Creating plot for velocity magnitude and pressure for probe points 
7
Traceback (most recent call last):
 line 514, in plot_probe_points
    fig.delaxes(axs[i])
IndexError: index 7 is out of bounds for axis 0 with size 4
```

keiyamamo: If seems like `axs` needs to be flattened? It seems to be very similar to `plot_probe_points_tke` but there’s slight difference how `axs` and `axes` are used.

keiyamamo:
@johannesring  I tested this function locally and looks very nice:) I had one error related to removing empty plot and please see my comments.

src/fsipy/automatedPostprocessing/log_plotter.py
johannesring: Thanks, this should be fixed now.

johannesring:
@keiyamamo - If you use `--compute-average`, does it compute what you want?

johannesring:
@keiyamamo - Thanks for the info. I have tried to implement phase-averaging TKE now. Please let me know what you think.

keiyamamo (commented):
Hi @johannesring. Thank  you so much for the update! It is definitely closer to what I’m thinking but there still needs some modification. As a result of phase-averaging, we need to get mean velocity that has only one cycle length and use that to compute fluctuating velocity for all the cycles. If it’s not clear, we can discuss at Simula :)

johannesring:
Thanks @keiyamamo, I have made another attempt now. Please check it out.

keiyamamo:
It looks great :) Thanks!

keiyamamo (approved):

Issue #69, opened 2023-10-02
fsipy requires `morphMan` for all the functionality 


keiyamamo:
https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/28422ae70158c4b6581ce0339308ae6ee25b6bcf/src/fsipy/__init__.py#L8-L10

I tried to use `fsipy-log-plotter` on the cluster and figured out that `morphMan` is required even when we just want to use post-processing tools. We could do same as `VaMPy` (use try: except method) but I do wonder if we even need this import in `__init__.py` under `src/`. It makes more sense to import functions but not really file itself. Hope to hear what you think! @johannesring @jorgensd

johannesring:
@keiyamamo - Can you check if pull request #70 fixes the issue on the cluster?

keiyamamo:
Hi @johannesring, 

Thanks for the fix! Yes, this fixes the issue. However, when I tried `fsipy-predeform-mesh`. I got the same error message due to `from . import automated_preprocessing` inside `src/fsipy/automatedPreprocessing/__init__.py`. Technically, you don’t need `morphMan` for using `fsipy-predeform-mesh`. Do you think it is possible to remove this error message or should we assume that the user has `morphMan` since `fsipy-predeform-mesh` is part of pre-processing?

johannesring:
@keiyamamo - One way to solve this would be to create subfolders for the mesh generation (`automated_preprocessing.py`, `preprocessing_common.py` and `vmtkmeshgeneratorfsi.py`) and for the predeform script. Another way would be to move the predeform script to the post processing folder. What do you think?

johannesring:
Fixed in PR #70.

PR #71, opened 2023-10-03
Move pre-deform mesh script

johannesring:
This PR moves the pre-deform mesh script to post-processing folder as discussed in issue #69.


keiyamamo (approved):
Thanks for quick implementation!

PR #76, opened 2023-10-05
Replace pickle with json for loading default variables

johannesring:
This pull request updates the code in `separate_mesh.py` to load data from a JSON file (`default_variables.json`) instead of a pickled file (`default_variables.pickle`). This to reflect the change in `turtleFSI`.


keiyamamo:
Thank you for the fix!

keiyamamo (approved):

PR #77, opened 2023-10-07
Refactor scaling

keiyamamo:
We had scaling from mm to m as a default when converting from xml to HDF5 while probes and other information inside `_info.json` file were not scaled. This PR changes default scaling to 1 (no scaling) and that the user should scale the input surface at the beginning of meshing so that probes are generated with scaled surface.

### DIFF
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -430,7 +430,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
                                                    solid_thickness,
                                                    solid_thickness_parameters)
         except Exception:
-            print("ERROR: Mesh generation failed. Try to remesh with alternative method.")
+            print("Trying to remesh with alternative method.")
             distance_to_sphere = mesh_alternative(distance_to_sphere)
             mesh, remeshed_surface = generate_mesh(distance_to_sphere,
                                                    number_of_sublayers_fluid,
@@ -764,7 +764,7 @@ def read_command_line(input_path=None):
                 mesh_format=args.mesh_format, flow_rate_factor=args.flow_rate_factor,
                 solid_side_wall_id=args.solid_side_wall_id, interface_fsi_id=args.interface_fsi_id,
                 solid_outer_wall_id=args.solid_outer_wall_id, fluid_volume_id=args.fluid_volume_id,
-                solid_volume_id=args.solid_volume_id)
+                solid_volume_id=args.solid_volume_id, verbosity=args.verbosity)
 
 
 def main_meshing():
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -123,6 +123,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
     # Scale surface
     if scale_factor is not None:
         surface = scale_surface(surface, scale_factor)
+        resampling_step *= scale_factor
 
     # Check if surface is closed and uncapps model if True
     is_capped = check_if_closed_surface(surface)
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -430,6 +430,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
                                                    solid_thickness,
                                                    solid_thickness_parameters)
         except Exception:
+            print("ERROR: Mesh generation failed. Try to remesh with alternative method.")
             distance_to_sphere = mesh_alternative(distance_to_sphere)
             mesh, remeshed_surface = generate_mesh(distance_to_sphere,
                                                    number_of_sublayers_fluid,
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -27,7 +27,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
                        meshing_method, refine_region, is_atrium, add_flow_extensions, visualize, config_path,
                        coarsening_factor, inlet_flow_extension_length, outlet_flow_extension_length,
                        number_of_sublayers_fluid, number_of_sublayers_solid, edge_length,
-                       region_points, compress_mesh, scale_factor, resampling_step, meshing_parameters,
+                       region_points, compress_mesh, scale_factor, scale_factor_h5, resampling_step, meshing_parameters,
                        remove_all, solid_thickness, solid_thickness_parameters, mesh_format, flow_rate_factor,
                        solid_side_wall_id, interface_fsi_id, solid_outer_wall_id, fluid_volume_id, solid_volume_id):
     """
@@ -453,7 +453,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
 
     if mesh_format == "hdf5":
         print("--- Converting XML mesh to HDF5\n")
-        convert_xml_mesh_to_hdf5(file_name_xml_mesh)
+        convert_xml_mesh_to_hdf5(file_name_xml_mesh, scale_factor_h5)
         # Evaluate edge length for inspection
         edge_length_evaluator(file_name_xml_mesh, file_name_edge_length_xdmf)
     elif mesh_format == "xdmf":
@@ -644,6 +644,12 @@ def read_command_line(input_path=None):
 
                         help="Scale input model by this factor. Used to scale model to [mm].")
 
+    parser.add_argument("-sch5", "--scale-factor-h5",
+                        default=1.0,
+                        type=float,
+                        help="Scaling factor for HDF5 mesh. Used to scale model to [mm]." +
+                             "Note that probes and other parameters are not scaled.")
+
     parser.add_argument('-rs', '--resampling-step',
                         default=0.1,
                         type=float,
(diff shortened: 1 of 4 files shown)


src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -123,6 +123,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
     # Scale surface
     if scale_factor is not None:
         surface = scale_surface(surface, scale_factor)
+        resampling_step *= scale_factor
```
johannesring, line 126: I think it would be good to add a note about this for the `--resampling-step` command line argument. For example, something like this:
``` python
parser.add_argument('-rs', '--resampling-step',
                    default=0.1,
                    type=float,
                    help="Resampling step used to resample centerline in [m]. "
                         "Note: If --scale-factor is used, this step will be adjusted accordingly.")
```
Also, I think we should scale the solid thickness similarly when using `--scale-factor`. We should handle both `constant` and `variable` solid thickness.

src/fsipy/automatedPreprocessing/preprocessing_common.py
```diff
@@ -166,13 +166,15 @@ def generate_mesh(surface: vtkPolyData, number_of_sublayers_fluid: int, number_o
...
+def convert_xml_mesh_to_hdf5(file_name_xml_mesh: str, scaling_factor: float = 1) -> None:
     """Converts an XML mesh to an HDF5 mesh.
 
     Args:
         file_name_xml_mesh (str): The name of the XML mesh file.
         scaling_factor (float, optional): A scaling factor to apply to the mesh coordinates.
-                                          The default value is 0.001, which converts from millimeters to meters.
+                                          The default value is 1 (no scaling). Note that probes
+                                          and parameters inside _info.json file will not be scaled
+                                          if you only scale HDF5 file.
```
johannesring, line 177: Should we have a command line argument for only scaling the HDF5 file?

src/fsipy/automatedPreprocessing/automated_preprocessing.py
keiyamamo: Agreed! I added note for resampling step as well as scaling of the solid thickness.

src/fsipy/automatedPreprocessing/preprocessing_common.py
keiyamamo: Yes, it’s better not to scale only HDF5 file, but it’s better to keep the functionality. I added a command-line argument for it.

src/fsipy/automatedPreprocessing/automated_preprocessing.py
johannesring: Thanks! I think that only the `offset` and `min` and `max` distances should be scaled by `scale_factor`, not the `scale` value. I purpose the following changes, but feel free to use what you want:
``` patch
diff --git a/src/fsipy/automatedPreprocessing/automated_preprocessing.py b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
index 531a21e..0ca3158 100755
--- a/src/fsipy/automatedPreprocessing/automated_preprocessing.py
+++ b/src/fsipy/automatedPreprocessing/automated_preprocessing.py
@@ -125,7 +125,6 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
         print(f"--- Scale model by factor {scale_factor}\n")
         surface = scale_surface(surface, scale_factor)
         resampling_step *= scale_factor
-        solid_thickness_parameters = [scale_factor * i for i in solid_thickness_parameters]
 
     # Check if surface is closed and uncapps model if True
     is_capped = check_if_closed_surface(surface)
@@ -389,11 +388,15 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
     if solid_thickness == 'variable':
         if not path.isfile(file_name_distance_to_sphere_solid_thickness):
             if len(solid_thickness_parameters) == 4:
-                distance_offset, distance_scale, min_distance, max_distance = solid_thickness_parameters
+                # Apply scale factor to offset, min distance, and max distance
+                if scale_factor is not None:
+                    solid_thickness_parameters[0] *= scale_factor  # Offset
+                    solid_thickness_parameters[2] *= scale_factor  # Min distance
+                    solid_thickness_parameters[3] *= scale_factor  # Max distance
+
                 distance_to_sphere = distance_to_spheres_solid_thickness(distance_to_sphere,
                                                                          file_name_distance_to_sphere_solid_thickness,
-                                                                         distance_offset, distance_scale,
-                                                                         min_distance, max_distance)
+                                                                         *solid_thickness_parameters)
             elif len(solid_thickness_parameters) == 0:
                 distance_to_sphere = distance_to_spheres_solid_thickness(distance_to_sphere,
                                                                          file_name_distance_to_sphere_solid_thickness)
@@ -408,6 +411,10 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
             print("ERROR: Invalid parameter for constant solid thickness. This should be a " +
                   "single number greater than zero.")
             sys.exit(-1)
+        else:
+            # Apply scale factor to the constant thickness value
+            if scale_factor is not None:
+                solid_thickness_parameters[0] *= scale_factor
 
     # Compute mesh
     if not path.isfile(file_name_vtu_mesh):
@@ -647,13 +654,14 @@ def read_command_line(input_path=None):
     parser.add_argument("-sch5", "--scale-factor-h5",
                         default=1.0,
                         type=float,
-                        help="Scaling factor for HDF5 mesh. Used to scale model to [mm]." +
-                             "Note that probes and other parameters are not scaled.")
+                        help="Scaling factor for HDF5 mesh. Used to scale model to [mm]. " +
+                             "Note that probes and other parameters are not scaled. " +
+                             "Do not use in combination with --scale-factor.")
 
     parser.add_argument('-rs', '--resampling-step',
                         default=0.1,
                         type=float,
-                        help="Resampling step used to resample centerline in [m]." +
+                        help="Resampling step used to resample centerline in [m]. " +
                              "Note: If --scale-factor is used, this step will be adjusted accordingly.")
 
     parser.add_argument('-mp', '--meshing-parameters',
@@ -682,11 +690,13 @@ def read_command_line(input_path=None):
                         type=float,
                         nargs="+",
                         default=[0.3],
-                        help="Parameters for solid thickness [m]. For constant solid thickness, this should be " +
-                             "given as a single float. For 'variable' solid thickness, this should be given as " +
-                             "four floats for the distancetosphere scaling function: 'offset', 'scale', 'min' " +
-                             "and 'max'. For example --solid-thickness-parameters 0 0.1 0.25 0.3 " +
-                             "Note: If --scale-factor is used, solid thickness will be adjusted accordingly.")
+                        help="Parameters for solid thickness [m]. For 'constant' solid thickness, provide a single " +
+                             "float. For 'variable' solid thickness, provide four floats for the distancetosphere " +
+                             "scaling function: 'offset', 'scale', 'min' and 'max'. " +
+                             "For example --solid-thickness-parameters 0 0.1 0.25 0.3. " +
+                             "Note: If --scale-factor is used, 'offset', 'min', and 'max' parameters will be " +
+                             "adjusted accordingly for 'variable' solid thickness, and the constant value will also " +
+                             "be scaled for 'constant' solid thickness.")
 
     parser.add_argument('-mf', '--mesh-format',
                         type=str,

```

src/fsipy/automatedPreprocessing/preprocessing_common.py
johannesring: I added a note that this new command line argument should not be used together with `--scale-factor` (see diff posted above)."

keiyamamo:
Hi @johannesring 

Thanks for the suggestion! I updated the script based on your suggestion, but the test became unstable (sometimes it passes and sometimes it doesn’t). The problematic test is `test_mesh_model_with_variable_mesh_density`. Do you think we need to also scale `meshing_parameters` like we did for `solid-thickness-parameters`?

keiyamamo:
Yes, it seems like it's related to #73. I reverted the commit and now the test passes all the time.

johannesring:
> Yes, it seems like it's related to #73. I reverted the commit and now the test passes all the time.

Ok, but I don't think the test is useful. If you look at `mesh.GetNumberOfPoints()` and `remeshed_surface.GetNumberOfPoints()` before and after the test, you can see that they are exactly the same. So, it basically just runs the mesh generation twice.

keiyamamo:
I see. Actually, I remember talking with Henrik about the random error from tetgen and I think this re-meshing was his strategy to avoid the random error. So, I agree with you that it runs meshing twice but it reduces the chance of getting tetgen related random error on my machine. I’ll experiment a bit to confirm if this is actually the case and will get back to you.

johannesring:
I agree that re-meshing is good when there is a random error from tetgen, but the test `if mesh.GetNumberOfPoints() < remeshed_surface.GetNumberOfPoints()` is not a good test to determine this. At least not in this script. Maybe it works in VaMPy. 

A better way to handle the random error might be to raise an exception in [this](https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/ef70005c0a5076c1736a272ac7f8a1be99f90ee0/src/fsipy/automatedPreprocessing/vmtkmeshgeneratorfsi.py#L373) line instead of just printing to the screen. Then [this](https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/ef70005c0a5076c1736a272ac7f8a1be99f90ee0/src/fsipy/automatedPreprocessing/automated_preprocessing.py#L412-L424) `try`-`except` will handle this.

Please go ahead and experiment and let me know what you find.

keiyamamo:
Hi @johannesring 

After experimenting, I agree with you that the test `if mesh.GetNumberOfPoints() < remeshed_surface.GetNumberOfPoints()`does not seem useful and it’s better to remove it. I changed from just printing to raise exception and also added `tetgen.Verbose = 1`. With this we get more direct information from tetgen and I think it would be convenient to have it in  some cases. 

With this, scaling should work fine now.

src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -416,6 +430,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
                                                    solid_thickness,
                                                    solid_thickness_parameters)
         except Exception:
+            print("ERROR: Mesh generation failed. Try to remesh with alternative method.")
```
johannesring, line 433: Maybe "Trying to remesh with alternative method."?

johannesring (approved):
Looks good! Maybe `tetgen.Verbose` should only be turned on when running `fsipy-mesh` with `-v`?

src/fsipy/automatedPreprocessing/automated_preprocessing.py
```diff
@@ -430,7 +430,7 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
                                                    solid_thickness,
                                                    solid_thickness_parameters)
         except Exception:
-            print("ERROR: Mesh generation failed. Try to remesh with alternative method.")
+            print("Trying to remesh with alternative method.")
```
johannesring, line 433: Sorry, I only meant to change the last part of the message. I think this is fine:
``` python
print("ERROR: Mesh generation failed. Trying to remesh with alternative method.")
```
Simply replace `Try` with `Trying`.

keiyamamo:
> Looks good! Maybe `tetgen.Verbose` should only be turned on when running `fsipy-mesh` with `-v`?

I tried so that `tegen.Verbose` is only activated with `-v` flag, but somehow tetgen meshing became unstable. Have no idea why but I decided to remove `tetgen.Verbose` since it is not essential.

PR #81, opened 2023-10-17
improve offset stenosis problem file

keiyamamo:
This PR improves offset stenosis problem file. Now it includes functionality from `simulation_common.py`for printing CFL, Reynolds number, velocity/pressure from probes.


johannesring (approved):
Looks good!

PR #82, opened 2023-10-17
Add aneurysm simulation file

dbruneau-mie:
Hey Kei, here is the un-tested aneurysm simulation file. Attatching the aneurysm mesh file here as a .txt file, need to change the extension to .h5 to use it.
[file_aneurysm.txt](https://github.com/KVSlab/Aneurysm_Workflow_FSI/files/12932577/file_aneurysm.txt)
For permanent storage/linking these mesh files, should we upload the meshes as separate entries in Zenodo, or upload all of them in a single folder on Zenodo?

### DIFF
--- a/src/fsipy/simulations/aneurysm.py
+++ b/src/fsipy/simulations/aneurysm.py
@@ -1,3 +1,216 @@
 """
 Problem file for aneurysm FSI simulation
 """
+import os
+import numpy as np
+
+from vampy.simulation.Womersley import make_womersley_bcs, compute_boundary_geometry_acrn
+from turtleFSI.problems import *
+from dolfin import HDF5File, Mesh, MeshFunction, facets, assemble, UserExpression, sqrt, FacetNormal, ds, \
+    DirichletBC, Measure, inner, parameters
+
+# set compiler arguments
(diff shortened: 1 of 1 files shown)


keiyamamo:
Hi David, 

Thanks! I think we can have them as separate entries in zenodo. I tried to set up zenodo but figured out that I need to put quite many details about the project (funding, authors and etc..), so thought it might be better to wait and do it when the repository is nearly complete. But, perhaps we can just set up zenodo and fill out the details later, I don’t really know how zenodo works..

dbruneau-mie:
> Hi David,
> 
> Thanks! I think we can have them as separate entries in zenodo. I tried to set up zenodo but figured out that I need to put quite many details about the project (funding, authors and etc..), so thought it might be better to wait and do it when the repository is nearly complete. But, perhaps we can just set up zenodo and fill out the details later, I don’t really know how zenodo works..

I will just create an entry for each mesh, fill in what details I can, then we can just create a new Zenodo entry later if it needs to be updated. I think the purpose of Zenodo is to archive data in a permanent way, changing the entry may defeat the purpose.

keiyamamo (approved):

PR #87, opened 2023-10-25
Retry mesh generation on TetGen failures

johannesring:
This PR adds an option `--mesh-generation-retries` to control how many mesh generation retries to run on TetGen failures.


keiyamamo:
Hi @johannesring, 

I can see that `VaMPy` was updated and caused CI to fail. Do you think it’s something you can integrate in this PR or should we make another one specifically for `VaMPy` update?

johannesring:
@keiyamamo - I will take a look and see if it can be integrated into this PR.

keiyamamo (approved):

PR #89, opened 2023-10-31
update compute_readable_h5.py

keiyamamo:
This is a re-implementation of existing [compute_readable_h5.py](https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/master/postprocessing_fenics/compute_readable_h5.py)

The purpose of this script is the same as before, creating HDF5 file that can be read with FEniCS for other post-processing purposes such as stress, strain, and hemodynamics computation. 

There is a high chance that these newly added scripts will be updated as we work more on other FEniCS related script, but this script itself is usable based on preliminary test of computing WSS and solid stresses DAB’s script.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ fsipy-predeform-mesh = "fsipy.automatedPostprocessing.predeform_mesh:main"
 fsipy-refine-mesh = "fsipy.automatedPostprocessing.postprocessing_mesh.create_refined_mesh:main"
 fsipy-separate-mesh = "fsipy.automatedPostprocessing.postprocessing_mesh.separate_mesh:main"
 fsipy-log-plotter = "fsipy.automatedPostprocessing.log_plotter:main"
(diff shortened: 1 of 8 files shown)


johannesring (commented):
Looks fine. I have a few comments in addition to the minor inline comments:
* When running the script with default arguments on a simulation with five cycles, the script does not stop at `0.951`, which is set as the default end time. Removing the `try`-`except` in the `while`-loop and adding a `else` clause with a `break` as suggested, fixes this problem.
* Should the default end time be perhaps the end time for the simulation instead of `0.951`?
* There is a mix of using `logging` and `print` statements. Maybe use `logging` throughout the script?

keiyamamo:
Hi @johannesring  Thank you so much for the comments and suggestions! I incorporated most of the suggestions. Regarding the while loop, I agree that the current while loop needs some more work to make it robust. I will work on that tomorrow and will let you know when it is ready for review again.

keiyamamo:
Hi @johannesring 
It is ready for a review again! Please let me know if you have suggestions.

johannesring (approved):
Looks good now and works well. Thanks!

PR #96, opened 2023-11-16
Allow multiple domains when separating mesh into fluid and solid

### LINKED ISSUE #85: SubMesh only works for single marker
Hi @jorgensd 

I have a question specifically to you. In our framework, we separate the domain after simulation has completed using `SubMesh` function. So far, I’ve been using only single fluid/solid domain but we also use multiple fluids/solids. Base on the discussion [here](https://fenicsproject.org/qa/513/submesh-of-two-subdomains/), it seems not possible to define multiple domain using SubMesh with different markers (In the link, they enforce the different subdomains to have single marker?). 

So my question is : 
Is it possible to separate domains while keeping multiple markers for specific domain ? I also attached illustration where I try to explain what I want to achieve. Maybe is this something `MeshView` can handle?

https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/c6b8b5aa480417abe6ec4f9794ca354bc4b0830b/src/fsipy/automatedPostprocessing/postprocessing_mesh/separate_mesh.py#L39

Screenshot 2023-10-23 at 16 04 49

keiyamamo:
Fix #85 and also include minor update of other common scripts.

With this update, we can use multiple fluid/solid domain and keep these ID even after separating the whole mesh into sub mesh.


src/fsipy/automatedPostprocessing/postprocessing_fenics/create_hdf5.py
```diff
@@ -267,16 +274,13 @@ def main() -> None:
...
-            solid_domain_id = solid_domain_id[0]
-            logging.info("solid_domain_id is not int, using first element of list \n")
-
         logging.info(f"--- Fluid domain ID: {fluid_domain_id} and Solid domain ID: {solid_domain_id} \n")
 
-    if save_deg == 2:
+    if args.mesh_path:
+        mesh_path = Path(args.mesh_path)
+        logging.info("--- Using user-defined mesh \n")
+        assert mesh_path.exists(), "Mesh file not found."
```
johannesring, line 282: It would be useful to print the mesh path here:
``` python
        assert mesh_path.exists(), f"Mesh file '{mesh_path}' not found."
```

src/fsipy/automatedPostprocessing/postprocessing_fenics/postprocessing_fenics_common.py
```diff
@@ -14,6 +14,8 @@ def parse_arguments() -> argparse.Namespace:
     """
     parser = argparse.ArgumentParser()
     parser.add_argument("--folder", type=Path, help="Path to simulation results")
+    parser.add_argument('--mesh-path', type=str, default=None,
```
johannesring, line 17: This could also be `type=Path`?

johannesring (approved):
Looks good.

keiyamamo:
@johannesring  Thanks for the comments! Your suggestions are incorporated now.

PR #98, opened 2023-11-17
add test for stenosis problem

keiyamamo:
add pytest for stenosis problem file. I tried to generate mesh during the test instead of using pre-generated h5 file, but I kept getting different result between running it local and on github. Even thought the test was passing on my machine, it kept failing on github. I’m not really sure why it is different but it could be due to some slight difference in the resulting mesh. 

In the end, I decided to use pre-generated h5 file and probe points as inputs. Let me know what you think.


johannesring (approved):
Looks good.

Did you try to increase the tolerance in `np.isclose` when you generated the mesh during the test? If that doesn't help, then I am fine with having the generated mesh in the repository, but maybe it would be good to add a comment with the command you used to generate the mesh?

keiyamamo:
Hi @johannesring ,

I did not change the tolerance because the value was quite different when I printed out the value. Mesh might be different when generated locally and on github, but at least I could not detect any difference when I used `print_mesh_information` function.

keiyamamo:
```
[-0.008744301657286097, -3.9486130816023894e-05, -3.758327695654815e-05]
[-0.008626092761948598, 1.277556859105341e-05, -1.3079934294778602e-05]
```
The first line is expected result that had no problem running locally on my Mac and the second line was what I got when running on github. So the value is noticeably different.

johannesring:
I see. This pass if setting `atol` to `1e-03`:
``` python
In [17]: expected =[-0.008744301657286097, -3.9486130816023894e-05, -3.758327695654815e-05]

In [18]: actual = [-0.008626092761948598, 1.277556859105341e-05, -1.3079934294778602e-05]

In [19]: np.isclose(expected, actual, atol=1e-03)
Out[19]: array([ True,  True,  True])

In [20]: np.isclose(expected, actual, atol=1e-04)
Out[20]: array([False,  True,  True])
```
I am fine with using the generated mesh.

keiyamamo:
Yes, that makes sense since the value itself is smaller than tolerance (1e-05). I’ll merge this and we can potentially come back to this and try to generate mesh during the pytest.

PR #101, opened 2023-11-20
add separate domain visualization script

keiyamamo:
This PR adds functionality to get separate domain visualization for the velocity on fluid domain and displacement on either solid or entire domain. The outputted visualization depends on how you run `create_hdf5.py` prior to running this new script called `create_separate_domain_visualization.py`. 

Please see the following image as an example of displacement on solid domain only and velocity on fluid domain. 
Screenshot 2023-11-20 at 15 53 42


keiyamamo:
I realized that there is no point creating displacement visualization for the whole domain since we already have it from turtle. So this last change was made so that visualization is created only when .h5 file was created with solid mesh only. Sorry for making a lot of commits!

johannesring (approved):
Looks fine. Only some minor comments:

* Should you add this script to `pyproject.toml` such that one can run for instance `fsipy-create-separate-domain-viz`from the command lline?
*  It would be nice to see some information at the end about which files are produced by the script.

keiyamamo:
@johannesring  Thanks for the comment! I updated the script and also added folder named `Visualization_separate_domain`.

PR #103, opened 2023-11-21
Add spectrogram scripts

johannesring:
This PR adds the following scripts:
* `fsipy-create-spectrograms-chromagrams`: script for creating spectrograms and chromagrams
* `fsipy-create-spectrogram-from-components`: script for creating a combined spectrogram by averaging three component-wise spectrograms.
* `fsipy-create-spectrum`: script for creating a power spectrum plot


src/fsipy/automatedPostprocessing/postprocessing_h5py/create_spectrograms_chromagrams.py
```diff
@@ -0,0 +1,232 @@
...
+    # Create logger and set log level
+    logging.basicConfig(level=args.log_level, format="%(message)s")
+
+    # Create or read in spectrogram dataframe
+    dvp, df, case_name, image_folder, visualization_hi_pass_folder = \
+        spec.read_spectrogram_data(args.folder, args.mesh_path, args.save_deg, args.stride, args.start_time,
+                                   args.end_time, args.n_samples, args.ylim, args.sampling_region,
+                                   args.fluid_sampling_domain_id, args.solid_sampling_domain_id, args.r_sphere,
+                                   args.x_sphere, args.y_sphere, args.z_sphere, args.dvp, args.interface_only,
+                                   args.component, args.point_id, sampling_method=args.sampling_method)
```
keiyamamo, line 216: Would it be possible to remove some of the command-line argument and use value from `default_variables.json` instead?

src/fsipy/automatedPostprocessing/postprocessing_h5py/postprocessing_common_h5py.py
```diff
@@ -0,0 +1,470 @@
...
+
+    Args:
+        out_file (Union[str, Path]): Path to the output file.
+
+    Returns:
+        tuple: Tuple containing the surface topology and coordinates.
+    """
+    mesh = h5py.File(out_file, "r")
+    topology = mesh["Mesh/0/mesh/topology"][:, :]
+    coords = mesh["Mesh/0/mesh/geometry"][:, :]
```
keiyamamo, line 96: Is it safer to call `mesh.close()`? or maybe not necessary?

src/fsipy/automatedPostprocessing/postprocessing_h5py/postprocessing_common_h5py.py
```diff
@@ -0,0 +1,470 @@
...
+            all_ids: Node IDs for all elements, sorted in ascending order.
+    """
+    fluid_topology, wall_topology, all_topology = get_domain_topology(mesh_path)
+
+    # Find the unique node ids in the wall, fluid, and all topologies, sorted in ascending order
+    wall_ids = np.unique(wall_topology)
+    fluid_ids = np.unique(fluid_topology)
+    all_ids = np.unique(all_topology)
+
+    return fluid_ids, wall_ids, all_ids
```
keiyamamo, line 152: I implemented a function for getting IDs. This new version is basically combined version of old `get_domain_topology` and `get_domain_ids`. Do you think it is worth moving new function so that it can be used in `postprocessing_h5py` as well?

src/fsipy/automatedPostprocessing/postprocessing_h5py/postprocessing_common_h5py.py
```diff
@@ -0,0 +1,470 @@
...
+    Returns:
+        tuple: Tuple containing the surface topology and coordinates.
+    """
+    mesh = h5py.File(out_file, "r")
+    topology = mesh["Mesh/0/mesh/topology"][:, :]
+    coords = mesh["Mesh/0/mesh/geometry"][:, :]
+    return topology, coords
+
+
+def get_domain_topology(mesh_path: Union[str, Path]) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
```
keiyamamo, line 100: Please see the comment below. This function does not work with multiple fluid domain and also assumes fluid domain ID to be 1 and solid domain to be 2 or larger than that, which can be dangerous.

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,778 @@
...
+    parser.add_argument('--x-sphere', type=float, default=0.0,
+                        help="X-coordinate of the center of the sphere used to include points for spectrogram (in "
+                             "meters).")
+    parser.add_argument('--y-sphere', type=float, default=0.0,
+                        help="Y-coordinate of the center of the sphere used to include points for spectrogram (in "
+                             "meters).")
+    parser.add_argument('--z-sphere', type=float, default=0.0,
+                        help="Z-coordinate of the center of the sphere used to include points for spectrogram (in "
+                             "meters).")
+    parser.add_argument('--dvp', type=str, default="v",
```
keiyamamo, line 74: I would like `dvp` to be more general since it can technically be `wss` or other variables that are not included `dvp`. Something like `-voi, --variable-of-interest` is more intuitive to me.

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,778 @@
...
+    parser.add_argument('--sampling-method', type=str, default="RandomPoint",
+                        help="Sampling method for spectrogram generation. Choose from 'RandomPoint' (random nodes), "
+                             "'SinglePoint' (single point specified by 'point_id'), or 'Spatial' (ensures uniform "
+                             "spatial sampling, e.g., in the case of fluid boundary layer, the sampling will not bias "
+                             "towards the boundary layer).")
+    parser.add_argument('--n-samples', type=int, default=10000,
+                        help="Number of samples to generate spectrogram data (ignored for SinglePoint sampling).")
+    parser.add_argument('--point-id', type=int, default=-1000000,
+                        help="Point ID for SinglePoint sampling. Ignored for other sampling methods.")
+
```
keiyamamo, line 98: This empty line could be removed?

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,778 @@
...
+    parser.add_argument('--overlap-frac', type=float, default=0.75,
+                        help="Fraction of overlap between adjacent windows.")
+    parser.add_argument('--window', type=str, default="blackmanharris",
+                        help="Window function to be used for spectrogram computation.")
```
keiyamamo, line 102: Not important, but we could explanation of what other window functions are available. Here https://docs.scipy.org/doc/scipy/reference/signal.windows.html scipy lists possible window functions, so it might be nice to add this as a reference point.

keiyamamo (commented):
Thank you so much for the update! It looks great! I put comments by simply looking at the code, but I’ll try this myself and might come up with more suggestions. For now, my main suggestions are 

1. I would like to avoid having many command line arguments. I would say it is safe to read parameter from `default_variables.json` and use these parameters as much as possible to reduce the number of command-line arguments, such as domain ID or coordinate of sphere. 
2. There are actually some overlapping function from `postprocessing_fenics` folder. Maybe it is worth moving those functions under `automatedPostprocessing` ? What do you think?

Again, it is very nice work and thanks!

src/fsipy/automatedPostprocessing/postprocessing_h5py/postprocessing_common_h5py.py
keiyamamo: Here is the link to the function https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/d98e4c73b9136592df3b06214a1b31ee2ee31b58/src/fsipy/automatedPostprocessing/postprocessing_fenics/create_hdf5.py#L174-L210

src/fsipy/automatedPostprocessing/postprocessing_h5py/postprocessing_common_h5py.py
```diff
@@ -0,0 +1,470 @@
...
+        elif 'VisualisationVector' in line:
+            h5_pattern = '"HDF">(.+?):/'
+            h5_str = re.findall(h5_pattern, line)
+            h5_ts.append(h5_str[0])
+
+            index_pattern = "VisualisationVector/(.+?)</DataItem>"
+            index_str = re.findall(index_pattern, line)
+            index = int(index_str[0])
+            index_ts.append(index)
+
```
keiyamamo, line 327: This is also implemented as a function here 
https://github.com/KVSlab/Aneurysm_Workflow_FSI/blob/d98e4c73b9136592df3b06214a1b31ee2ee31b58/src/fsipy/automatedPostprocessing/postprocessing_fenics/create_hdf5.py#L213

johannesring:
@keiyamamo - Thank you for your comments! I agree that it would be nice to use the functions that are already in `postprocessing_fenics`. I will look into that and also try to use `default_parameters.json`.

keiyamamo:
@johannesring  Good to hear! I think those functions could be moved under `automatedPostprocessing` and have it as `postprocessing_common.py` or something like that. Please let me know if you need any help with it!

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
johannesring: @keiyamamo - How about `--quantity`:
``` python
    parser.add_argument('-q', '--quantity', type=str, default="v",
                        help="Quantity to postprocess. Choose 'v' for velocity, 'd' for displacement, 'p' for "
                             "pressure, or 'wss' for wall shear stress.")
```
Is that okay with you? Or do you prefer your suggestion?

keiyamamo: Quantity sounds good to me! Please go ahead with that.

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,747 @@
...
+    parser.add_argument('--num-windows-per-sec', type=int, default=4,
+                        help="Number of windows per second for spectrogram computation.")
+    parser.add_argument('--thresh-val', type=int, default=None,
+                        help="Threshold value for the spectrogram. Default is determined based on the 'quantity' "
+                             "argument: if 'd', default is -42; if 'v', default is -20; if 'p', default is -5; if "
+                             "'wss', default is -18.")
+    parser.add_argument('--max-plot', type=int, default=None,
+                        help="Maximum value for plotting the spectrogram. Default is determined based on the "
+                             "'quantity' argument: if 'd', default is -30; if 'v', default is -7; if 'p', default is "
+                             "5; if 'wss', default is 0.")
```
keiyamamo, line 106: `--thresh-val` and  `--max-plot` defines minimum and maximum value of color range when plotting the spectrogram. I think it would be nice to rename those two so that it is more clear what we mean by these two names. For example, `--min-color-value` `--max-color-value`. They are `vmin` and `vmax` in matplotlib document, respectively. https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.pcolormesh.html

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,747 @@
...
+    visualization_path = folder_path / "Visualization"
+
+    logging.info(f"--- Processing folder path {folder_path}\n")
+
+    mesh_name_suffix = "" if save_deg == 1 else "_refined"
+    mesh_path = Path(mesh_path)
+    mesh_path = mesh_path.with_name(f"{mesh_path.stem}{mesh_name_suffix}{mesh_path.suffix}")
+    mesh_path_fluid = mesh_path.with_name(f"{mesh_path.stem}_fluid.h5")  # Needed for formatting SPI data
+
+    formatted_data_folder_name = f"res_{case_name}_stride_{stride}t{start_t}_to_{end_t}save_deg_{save_deg}"
```
keiyamamo, line 191: This formatted data folder name is hard to read. I think we can remove `case_name` because it is clear what folder we are working on and maybe can keep other parameters. 

So for example, 
```suggestion
    formatted_data_folder_name = f”npz_{start_t}s_to_{end_t}s_stride_{stride}save_deg_{save_deg}"
```
Other possibility is to make folder name more general and just add json file containing what parameters were used to generate these npz files, but that can be too much to do.

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,747 @@
...
+
+    if quantity == "wss":
+        wss_output_file = visualization_separate_domain_folder / "WSS_ts.h5"
+        surface_elements, coords = get_surface_topology_coords(wss_output_file)
+    else:
+        coords = get_coords(mesh_path)
+
+    if sampling_region == "sphere":
+        # Get wall and fluid ID's
+        fluid_ids, wall_ids, all_ids = get_domain_ids(mesh_path, fluid_domain_id, solid_domain_id)
```
keiyamamo, line 235: `wall_ids` should be renamed as `solid_ids` to make is consistent to the naming of `solid_domain_id`
Subsequently, all the other `wall_ids` need to be renamed as well.,
```suggestion
        fluid_ids, solid_ids, all_ids = get_domain_ids(mesh_path, fluid_domain_id, solid_domain_id)
```

keiyamamo (commented):
Hi @johannesring ,

Thank you for the updates! I tested most of the aspects of the new spectrograms and they are working great. I particularly liked the progress bar since it is much more clear how the process is going. All the logging information I get is also very helpful. 

I think it is almost ready to be merged but some naming need modification so that it is easier to understand. 

In addition to my comments, we also discussed offline

- move command-line argument function under each script instead of having it as common. This helps user to understand which parameters are actually required to run for a particular script.
- use `formatter_class=argparse.ArgumentDefaultsHelpFormatter,` to display default value when running `--help`

johannesring:
@keiyamamo - Thank you for your comments! I think I have addressed all of them now.

keiyamamo:
Hi @johannesring, 
Thank you for the update! I will do final check on Saga with my aneurysm data and then will come back to you if I have any more suggestions :)

src/fsipy/automatedPostprocessing/postprocessing_h5py/spectrograms.py
```diff
@@ -0,0 +1,743 @@
...
+    visualization_path = folder_path / "Visualization"
+
+    logging.info(f"--- Processing folder path {folder_path}\n")
+
+    mesh_name_suffix = "" if save_deg == 1 else "_refined"
+    mesh_path = Path(mesh_path)
+    mesh_path = mesh_path.with_name(f"{mesh_path.stem}{mesh_name_suffix}{mesh_path.suffix}")
+    mesh_path_fluid = mesh_path.with_name(f"{mesh_path.stem}_fluid.h5")  # Needed for formatting SPI data
+
+    formatted_data_folder_name = f"npz_{start_t}s_to_{end_t}s_stride_{stride}save_deg_{save_deg}"
```
keiyamamo, line 187:
```suggestion
    formatted_data_folder_name = f"npz_{start_t}s_to_{end_t}s_stride_{stride}_save_deg_{save_deg}"
```

src/fsipy/automatedPostprocessing/postprocessing_h5py/create_spectrograms_chromagrams.py
```diff
@@ -0,0 +1,256 @@
...
+
+def main():
+    # Load in case-specific parameters
+    args = spec.read_command_line_spec()
+
+    # Create logger and set log level
+    logging.basicConfig(level=args.log_level, format="%(message)s")
+
+    # Load parameters from default_parameters.json
+    parameters = read_parameters_from_file(args.folder)
```
keiyamamo, line 223: Is it possible to add `assert` to check if the `args.folder` exist?

keiyamamo (approved):
Hi @johannesring ,

It worked well on Saga as well. I have last few suggestions. Those are mostly related to my personal preference and how keep the data on Saga, so no pressure for the implementation :) It was very nice to run and see that it works on Saga :)

src/fsipy/automatedPostprocessing/postprocessing_h5py/create_spectrograms_chromagrams.py
keiyamamo: I had wrong folder path but the error message came from not finding `.json` file so it is my stupid mistake but would be nice to check if the folder exist in the first place.

PR #106, opened 2023-11-27
 Add pre-deform simulation (and fix cylinder.py)

dbruneau-mie:
I have added a simulation to pre-deform meshes. IT is similar to the "cylinder.py" simulation but has options to shift the start and end time of the pressure and velocity ramp. Also, the cylinder.py wouldn't run because it was missing a couple improts, that is also fixed in this PR

### DIFF
--- a/src/fsipy/simulations/predeform.py
+++ b/src/fsipy/simulations/predeform.py
@@ -7,7 +7,7 @@ this deformation as an approximate "zero-pressure" geometry, using "predeform_me
 import numpy as np
 from turtleFSI.problems import *
 from dolfin import HDF5File, Mesh, MeshFunction, assemble, UserExpression, FacetNormal, ds, \
-    DirichletBC, Measure, inner, parameters, SpatialCoordinate, Constant
+    DirichletBC, Measure, inner, parameters, SpatialCoordinate, Constant, facets, sqrt
 
 # set compiler arguments
 parameters["form_compiler"]["quadrature_degree"] = 6
@@ -32,7 +32,7 @@ def set_problem_parameters(default_variables, **namespace):
             T=0.3,  # Simulation end time
             dt=0.001,  # Time step size
             theta=0.501,  # Theta scheme (implicit/explicit time stepping)
-            save_step=1,  # Save frequency of files for visualisation
+            save_step=10,  # Save frequency of files for visualisation
             checkpoint_step=50,  # Save frequency of checkpoint files
             # Linear solver parameters
             linear_solver="mumps",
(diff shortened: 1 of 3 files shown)


src/fsipy/simulations/predeform.py
```diff
@@ -0,0 +1,210 @@
...
+    mu_s_val = E_s_val / (2 * (1 + nu_s_val))  # 0.345E6
+    lambda_s_val = nu_s_val * 2.0 * mu_s_val / (1.0 - 2.0 * nu_s_val)
+
+    default_variables.update(
+        dict(
+            # Temporal parameters
+            T=0.3,  # Simulation end time
+            dt=0.001,  # Time step size
+            theta=0.501,  # Theta scheme (implicit/explicit time stepping)
+            save_step=1,  # Save frequency of files for visualisation
```
keiyamamo, line 35: `save_step` could be larger than 1 in case of pre-deformation?

src/fsipy/simulations/predeform.py
```diff
@@ -0,0 +1,210 @@
...
+            folder="predeform_results",  # output folder generated for simulation
+            save_deg=1,  # save_deg=1 saves corner nodes only, save_deg=2 saves corner + mid-point nodes for viz
+        )
+    )
+
+    return default_variables
+
+
+def get_mesh_domain_and_boundaries(mesh_path, folder, **namespace):
+    print("Obtaining mesh, domains and boundaries...")
```
keiyamamo, line 81: would be nice to print from only rank 0 in case of MPI

src/fsipy/simulations/predeform.py
```diff
@@ -0,0 +1,210 @@
...
+    mesh = Mesh()
+    hdf = HDF5File(mesh.mpi_comm(), mesh_path, "r")
+    hdf.read(mesh, "/mesh", False)
+    boundaries = MeshFunction("size_t", mesh, 2)
+    hdf.read(boundaries, "/boundaries")
+    domains = MeshFunction("size_t", mesh, 3)
+    hdf.read(domains, "/domains")
+
```
keiyamamo, line 89: In case of aneurysm, we would want to modify fsi region like we do in `aneurysm.py`. I think it’s worth adding that in here.

src/fsipy/simulations/predeform.py
```diff
@@ -0,0 +1,210 @@
...
+            print("P = {} Pa".format(self.P))
+
+    def eval(self, value, x):
+        value[0] = self.P
+
+    def value_shape(self):
+        return ()
+
+
+def create_bcs(dvp_, DVP, mesh, boundaries, t_start_v, t_end_v, t_start_p, t_end_p, P_final,
```
keiyamamo, line 170: `dvp_` is imported but not used in this function.

src/fsipy/simulations/predeform.py
```diff
@@ -0,0 +1,210 @@
...
+            extrapolation_sub_type="constant",  # ["constant","small_constant","volume","volume_change","bc1","bc2"]
+            folder="predeform_results",  # output folder generated for simulation
+            save_deg=1,  # save_deg=1 saves corner nodes only, save_deg=2 saves corner + mid-point nodes for viz
+        )
+    )
+
+    return default_variables
+
+
+def get_mesh_domain_and_boundaries(mesh_path, folder, **namespace):
```
keiyamamo, line 80: `folder` is imported but not used in the function

keiyamamo (commented):
Hi @dbruneau-mie 

Thanks for adding pre-deformation problem file! I have some comments but overall it looks very good.

keiyamamo:
Hi @dbruneau-mie 

I tested the updated version locally and it works well!

keiyamamo (approved):

PR #110, opened 2023-11-28
Fixes in spectrogram code

johannesring:
This pull request addresses two issues in the spectrogram code:
1. Fixes an issue with the transformed matrix filename (removed `case_name`).
2. Resolves a circular import issue by moving the `sonify_point` function to `spectrograms.py`.


johannesring:
This is a simple fix, so I am merging this without a review.

PR #111, opened 2023-12-01
add new script for computing hemodynamics indices

keiyamamo:
This PR adds new implementation of computing hemodynamics indices. Here is summary of what I’ve done and also changes from the previous implementation. There are two references that were useful for implementing new wall shear stress

https://fenicsproject.org/qa/13680/different-integration-depending-elemtype-working-example/
https://fenicsproject.org/qa/13629/incorrect-stress-computations-compared-analytical-solution/

- all the quantities are computed based on the implementation `VaMPy`. Naming and the definition of each indices are identical. 
- new class `Stress` does not contain pressure term compared to `VaMPy`. Pressure term is always defined as zero in `VaMPy` anyway and does not contribute to the final WSS because it disappears during the process of extracting tangential component of wall shear stress.
- new class `Stress` uses projection that are valid for `CG2` velocity, defined as a class `SurfaceProject`. This class is cheaper than built in FEniCS `project` function since we cache left hand side matrix A, which is constant.  
- use `DG` function space instead of `CG` function space. This is due to the fact that `grad(u)` where `u` is continuous function should be discontinuous function. For example, it is discussed [here](https://fenicsproject.discourse.group/t/why-use-dg-space-to-project-stress-strain/3768/2)
- All the indices are now defined on the entire domain. This is a limitation of using `DG` function space and fenics. `interpolate` does not work well with `DG` function space since there’s no longer one-to-one mapping from vertices to the degree of freedom. In the future, this can be optimized.
- use `write_checkpoint` instead of `write`. This is also related to the usage of `DG` function space. Naive `write` function from fenics interpolate `DG`  function to `CG1`. 
- add `fsipy-compute-hemo` command.


src/fsipy/automatedPostprocessing/postprocessing_fenics/compute_hemodynamics.py
```diff
@@ -0,0 +1,332 @@
...
+    parser.add_argument('--mesh-path', type=Path, default=None,
+                        help="Path to the mesh file. If not given (None), " +
+                             "it will assume that mesh is located <folder_path>/Mesh/mesh.h5)")
+    parser.add_argument("--stride", type=int, default=1, help="Save frequency of output data")
+    args = parser.parse_args()
+
+    return args
+
+
+class SurfaceProject:
```
jorgensd, line 41: Maybe change name to `SurfaceProjector`?

src/fsipy/automatedPostprocessing/postprocessing_fenics/postprocessing_fenics_common.py
```diff
@@ -24,3 +26,29 @@ def parse_arguments() -> argparse.Namespace:
...
+        dolfin.Function: Projected function
+    """
+    v = TestFunction(V)
+    u = TrialFunction(V)
+    a = inner(u, v) * dx
+    L = inner(f, v) * dx
+    u_ = Function(V)
+    solver = LocalSolver(a, L)
+    solver.factorize()
+    solver.solve_local_rhs(u_)
```
jorgensd, line 52: We might nee to use `solve_global_rhs` here, if the input `f` is not in `DG`.
We should probably also add a check to ensure that `V` is a DG space.

src/fsipy/automatedPostprocessing/postprocessing_fenics/compute_hemodynamics.py
jorgensd, line 1: Could we add some tests for this new functionality?

keiyamamo:
Changed to draft for now due to possible enhancement.

keiyamamo:
Hi @jorgensd and @johannesring 

This is again ready for reviews. There are some changes from the initial PR. 

1. WSS and other hemodynamic indices are now saved only on the surface thanks to the help from @jorgensd!
2. I added pytest for this script and I will elaborate a bit about the test. 
- First, I test only `TAWSS`, which is time-averaged wall shear stress while the output of the script contains other indices, such as Oscillatory shear index (OSI). These other indices are dependent on the TAWSS and therefore I think it is sufficient to check the value of TAWSS. 
- Second, I uploaded mesh and `u.h5` which contains velocity data of two consecutive time steps. The original idea was to simply use `Expression` from FEniCS but I figured out that it is not straightforward due to how we read the data in the script. I hope this is acceptable and we can optimize this at later stage. My priority is to finish all the necessary functionality and I can keep working on improving the repo over time.

jorgensd (approved):

PR #115, opened 2023-12-05
Update stress strain code

keiyamamo:
add new implementation of computing stress and strain. 

Changes are

- use `write_checkpoint` to visualize DG stress/strain properly (same argument for WSS)
- compute Green-Lagrange strain instead of infinitesimal strain
- read material parameters from `default_variables.json` generated from turtleFSI

It is draft for now since I need to add pytest. If anyone has good suggestion for testing stress/strain computation, I would be glad to know!


johannesring (approved):
There is a missing copyright statement in `compute_stress_strain.py`, otherwise it looks fine.

PR #130, opened 2024-01-19
add solid probe functionality

keiyamamo:
This PR adds following functionality to VaSP

- `generate_solid_probe.py` will generate probe points inside the solid domain. This script can be run after the main meshing script and the user can specify FSI region by x,y,z coordinate with minimum and maximum. Later, we can merge this into main meshing, but for now it is created as a separate script.

- add functions ‘load_solid_probe_points’ and ‘print_solid_probe_points’ inside `simulation_common.py`. This is basically the same functions for fluid probe but just solid version. 

- add functionality to plot the probed displacement inside `log_plotter.py’.

Also I modified the offset stenosis problem file and pytest for that.


keiyamamo:
I changed the way we evaluate functions with probe points. The evaluation was not correct when running parallel, so I fixed the issue. There are two points that I changed.

1. make sure that functions is not allowed to extrapolate when evaluating. We want only single processor that contains the probe point to be able to evaluate. This is reflected as 

```python
if v.get_allow_extrapolation():
   v.set_allow_extrapolation(False)
```
2. Do not evaluate function components wise. Instead directly evaluated function and then print out each component. With `sub`, I noticed that the value was not correct and also it is more computationally expensive. They are fixed as 
```python
u_eval = peval(v, point.tolist())
```

PR #134, opened 2024-02-04
Log plotter extract probe

keiyamamo:
add functionality to save probe data as `.pkl` file. This makes it possible to compare probe data across different simulations. 
What you need to do is to simply add `--save-probes` with either `--plot-probe-points` or `--plot-probe-points-displacement`.


johannesring (approved):
Looks fine.

PR #140, opened 2024-03-06
Meshing script improvements

johannesring:
This PR includes the following meshing scripts improvements:
* Use `configargparse` to allow to generate a mesh using a config file (use the `--config` option in `vasp-generate-mesh`).
* Use `pathlib` for paths.
* Increase the number of default retries for mesh generation failures (trying to reduce failures on github).


keiyamamo (approved):
Looks great!

PR #145, opened 2024-03-18
Reduce file size

keiyamamo:
We previously switched from using `write` to `write_checkpoint` when saving DG function values (strain/stress, WSS and so on). I noticed that h5 file generated with `write_checkpoint` saves mesh information for all the time steps even though the mesh is static. With this update, we will only save mesh information at the first time step when creating high pass filtered strain, which reduces file size significantly. 

Additionally, I also switched to using `libver=latest` when opening/creating h5 file. This seeps up the saving of arrays. 
https://docs.h5py.org/en/stable/high/file.html#version-bounding

### DIFF
--- a/src/vasp/automatedPostprocessing/postprocessing_h5py/create_hi_pass_viz.py
+++ b/src/vasp/automatedPostprocessing/postprocessing_h5py/create_hi_pass_viz.py
@@ -233,10 +233,11 @@ def create_hi_pass_viz(formatted_data_folder: Path, output_folder: Path, mesh_pa
 
         elif quantity == "strain":
             # first open dof_info (dict)
-            assert dof_info is not None
-            for name, data in dof_info.items():
-                dof_array = vector_data.create_dataset(f"{array_name}/{name}", data=data)
-                dof_array[:] = data
+            if idx == 0:
+                assert dof_info is not None
+                for name, data in dof_info.items():
+                    dof_array = vector_data.create_dataset(f"{array_name}/{name}", data=data)
+                    dof_array[:] = data
 
             v_array = np.zeros((int(n_cells_fsi * 4), 9))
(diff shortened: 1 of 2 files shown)


johannesring:
Hi @keiyamamo. I also see very big file sizes for `TrueStress.h5` and `GreenLagrangeStrain.h5`. They are both 733G. My simulation was run with a mesh with 200k cells and 9507 timesteps (2 cycles with `dt=0.0002`). Is this expected?

keiyamamo:
Yes, unfortunately, using `write_checkpoint` makes files very large due to two reasons. 
1. As I mentioned, mesh information is saved every time step.
2.  Since we now use DG, the number of DOF increases compared to using CG functions. 

Are you computing stress in the solid domain only or for the entire domain? When creating HDF5 file with `vasp-create-hdf5 `, you can use `--extract-solid-only`.

johannesring:
Thanks, I guess I am using the entire domain since I did not specify `--extract-solid-only`.  When do I want to use the entire domain?

keiyamamo:
Hi @johannesring Actually, 733G sounds a bit too large. When running `vasp-create-hdf5`, I usually specify time window that I’m interested in and use `--extract-solid-only` ( for example, `--start-time 0.951` `--end-time 1.902`). Then you can reduce file size.

johannesring (approved):

keiyamamo:
> Thanks, I guess I am using the entire domain since I did not specify `--extract-solid-only`. When do I want to use the entire domain?

I don’t really know why we need to compute for the entire domain to be honest. It was something that I inherited from DAB and I just kept it. Maybe we should make `--extract-solid-only` as default and have a flag `--entire-domain` or something similar instead. What do you think?

johannesring:
Yes, that sounds good to me.

PR #149, opened 2024-03-22
Allow multiple runs of 'distancetospheres' method

johannesring:
This pull request adds the ability to run the `distancetospheres` meshing method multiple times with distinct parameters for each run.

### DIFF
--- a/src/vasp/automatedPreprocessing/automated_preprocessing.py
+++ b/src/vasp/automatedPreprocessing/automated_preprocessing.py
@@ -388,16 +388,20 @@ def run_pre_processing(input_model, verbose_print, smoothing_method, smoothing_f
             distance_to_sphere = read_polydata(str(file_name_distance_to_sphere_diam))
     elif meshing_method == "distancetospheres":
         if not file_name_distance_to_sphere_spheres.is_file():
-            if len(meshing_parameters) == 4:
-                if scale_factor is not None:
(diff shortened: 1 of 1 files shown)


keiyamamo (approved):
Looks good! Is this useful when you want to have more than one sphere ? I’m not sure if I understood when I should use this functionality.

johannesring:
This is useful for example if you want to refine the mesh in one region and then you want an even finer mesh at some critical points inside this region. For example, here I use 8 parameters for the `distancetospheres` meshing method:
```  console
vasp-generate-mesh -i C0002.vtp -m distancetospheres -mp 0 0.1 0.6 0.9 0 0.2 0.35 0.9
```
When running this, a window will pop up and I can place spheres where I want the finer mesh quality. The first four values given to the `-mp` option will be used, resulting in a mesh with edge length varying from 0.6 to 0.9. For example like this:
Screenshot from 2024-03-25 09-37-57
Screenshot from 2024-03-25 09-38-04
After closing the Window, a new window will appear. The result from the previous step will be used. Now, the next four values will be used, and I can place new spheres, for example like this:
Screenshot from 2024-03-25 09-38-48
The resulting mesh will look like this:
image
Hope this help you to understand how to use this.

keiyamamo:
@johannesring Thank you for the explanation! It is clear to me:)

PR #150, opened 2024-03-25
Add option to switch between 'euclidean' and 'geodesic' distance methods

johannesring:
Added the ability to switch between `euclidean` and `geodesic` distance computation methods in the preprocessing pipeline. Currently, this will only have effect when using the `distancetospheres` meshing method.

Here is an example of the difference between the two methods:
* `geodesic`
Screenshot from 2024-03-25 15-15-41
* `euclidean`
image


keiyamamo (approved):
Looks nice! Is there necessity/plan to add this functionality to othermeshing method? Just asking out of curiosity :)

johannesring:
@keiyamamo  - There is a PR for this in VaMPy: [PR #132](https://github.com/KVSlab/VaMPy/pull/132).

PR #154, opened 2024-03-27
add pytest for `create_hdf5.py`

keiyamamo:
add pytest for creating FEniCS compatible HDF5 files for the velocity and displacement after running FSI simulation.


johannesring (approved):
Looks good.

PR #158, opened 2024-04-12
Fix variable name conflict in meshing parameters loop

johannesring:
This pull request addresses a variable name conflict in the code related to meshing parameters.


johannesring:
This is a small fix, so I will merge this without a review.

PR #160, opened 2024-04-23
Fix issue with hemodynamic indices computation

### LINKED ISSUE #157: Bug  in computing hemodynamic indicies
I found a bug in hemodynamic indices computation. Here is the summary of the problem and how to fix. I will check if the new computation is correct and will make a PR once it is confirmed. 

**Source of the problem**
https://github.com/KVSlab/VaSP/blob/1e018b83d0922bb10206f10fdc5e34fd1b23dad9/src/vasp/automatedPostprocessing/postprocessing_fenics/compute_hemodynamics.py#L299

When computing the magnitude of wall shear stress, we use FEniCS `project` function. However, according to [this post](https://fenicsproject.discourse.group/t/dot-grad-u-grad-u-has-negative-values/1935), this so-called L2 projection does not preserve the sign and can cause oscillations. If I understand correctly, this problem is called [Gibbs Phenomenon](https://arxiv.org/pdf/1907.03429.pdf). 
In our case, I found that the `tawss` (or magnitude of wall shear stress) has negative values even though they should be positive. 

**How to fix**
Again, according to [this post](https://fenicsproject.discourse.group/t/dot-grad-u-grad-u-has-negative-values/1935), the fix is to use so-called lumped-projection (see below). I tested this lumped projection with simple stenosis case and one of the aneurysms, and found that all the values are positive and `OSI`is bounded to be between 0 - 0.5, as it should. Thus, I think this is the way to fix . Perhaps @jorgensd  can teach us this lumped projection ?

```python
def lumpedProject(f):
    v = TestFunction(V)
    lhs = assemble(inner(Constant(1.0),v)*dx)
    rhs = assemble(inner(f,v)*dx)
    u = Function(V)
    as_backend_type(u.vector())\
        .vec().pointwiseDivide(as_backend_type(rhs).vec(),
                               as_backend_type(lhs).vec())
    return u
```

keiyamamo:
Fix #157 and improved pytest to make sure that OSI value is between 0 and 0.5.

### DIFF
--- a/src/vasp/automatedPostprocessing/postprocessing_fenics/compute_hemodynamics.py
+++ b/src/vasp/automatedPostprocessing/postprocessing_fenics/compute_hemodynamics.py
@@ -14,14 +14,15 @@ from pathlib import Path
 import argparse
 
 from dolfin import Mesh, HDF5File, VectorFunctionSpace, Function, MPI, parameters, XDMFFile, TrialFunction, \
-    TestFunction, inner, ds, assemble, FacetNormal, sym, project, FunctionSpace, PETScDMCollection, grad, \
-    LUSolver, FunctionAssigner, BoundaryMesh, sqrt, Constant, dx, as_backend_type, IndexMap, assign
+    TestFunction, inner, ds, assemble, FacetNormal, sym, FunctionSpace, PETScDMCollection, grad, \
+    LUSolver, FunctionAssigner, BoundaryMesh
 
 from vampy.automatedPostprocessing.postprocessing_common import get_dataset_names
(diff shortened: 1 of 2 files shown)


johannesring (approved):
Looks good.

jorgensd:
We decided not to project the `sqrt(inner(tau, tau))` into `V_boundary` as it is not in a polynomial. (sqrt(DG-2) is not in  DG-1). This can cause slightly negative values.
We instead compute the magnitudes of the stress directly at the degrees of freedom (similar to what one would see in Paraview).

PR #162, opened 2024-05-07
fix problem with interface_only option

keiyamamo:
found a minor bug when selecting `--interface-only` option and is fixed. Also modified the `save_deg` parameter so that is is automatically set to 1 when selecting pressure.


keiyamamo:
Because it is a simple fix, I will merge without review.

Issue #167, opened 2024-07-15
CI fails due to typing issue with mypy


keiyamamo:
CI has been failing with the following error message

```
src/vasp/automatedPostprocessing/log_plotter.py:1026: error: Value of type "Axes" is not indexable  [index]
src/vasp/automatedPostprocessing/log_plotter.py:1035: error: Value of type "Axes" is not indexable  [index]
```

I found some reference that might be helpful: https://github.com/pydata/xarray/pull/9234

PR #168, opened 2024-07-16
Add type assertions and casts to fix mypy issues

johannesring:
This pull request addresses several issues reported by mypy by adding necessary type assertions and casts. Fixes issue #167.


keiyamamo (approved):
Looks great, thank you for the fix!

PR #176, opened 2024-08-30
Add functionality to check mesh inlets and outlets and flatten if necessary

dbruneau-mie:
Added new script "check_flatten_in_out.py", which takes a .h5 mesh file and checks whether inlets and outlets are flat, then flattens them if they fall below a certain tolerance

### DIFF
--- a/src/vasp/automatedPreprocessing/check_flatten_in_out.py
+++ b/src/vasp/automatedPreprocessing/check_flatten_in_out.py
@@ -1,0 +1,144 @@
+import argparse
+import h5py
+from pathlib import Path
+from shutil import copyfile
+import numpy as np
+import os
+
+
+def parse_arguments() -> argparse.Namespace:
+    """
+    Parse command line arguments.
+
+    Returns:
+        argparse.Namespace: Parsed command-line arguments.
+    """
+    parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


keiyamamo:
Hi @dbruneau-mie 

Thank you for the script! I tried with several geometries and can see that it improves the mesh.

Just one small comment is that solid inlet and outlet is still not flat. I guess this is not really an issue since flat outlet is important for only fluid domain. I was trying to figure out why inlet/outlet become non-flat and found out that it is due to BoundarylayerMesh from VMTK. In short, VMTK computes surface normal and use those vectors to create boundary layer mesh. However, those vectors are skewed at the ends, and cause this problem, as solid domain is constructed with boundary layer mesh. 

Screenshot 2024-09-02 at 10 38 48

It is possible to simply merge this, but it makes more sense to incorporate this as a final step of meshing. If it is okay with you, I will copy this function inside `src/vasp/automatedPreprocessing/preprocessing_common.py` and call it inside main meshing script. Let me know what you think.

dbruneau-mie:
Hey Kei, could you merge the PR, then integrate the function into the big
preprocessing script in a separate commit? I would like the isolated script
preserved at some moment in time, and that way I have my name on the commit

keiyamamo:
@dbruneau-mie Sure! I will merge and then create a separate commit.

keiyamamo (approved):

PR #189, opened 2024-09-18
move InnerP to simulation_common and clean terminology

keiyamamo:
`Class InnerP` was defined in several problem files, so decided to move it to `simulation_common.py`. At the same time, I cleaned up the terminology (now it is called InterfacePressure to be more specific).


keiyamamo:
Hi @johannesring 

I get an error from flake8
```
./src/vasp/simulations/offset_stenosis.py:9:1: F401 'turtleFSI.problems.*' imported but unused
```
but don’t really know what is the problem. I do the same import for `aneurysm.py` and other problem files, but only `offset_stenosis.py` with this PR gives this error. Do you have any idea how to fix this issue? 
I could add `# noqa: F401` and suppress this warning, but it would be nice not to do it.

Best,
Kei

keiyamamo:
In addition to the previous change, I added some more changes. 

1. removed `offset_stenosis_box.py`. This problem was essentially the same as the original `offset_stenosis.py` with a different way of defining FSI region. I simply added a note inside `offset_stenosis.py`, explaining how to use a box for choosing FSI region.
2. When reading Fourier coefficients, we previously used `os.path`, but switched to `pathlib`. This makes the code look cleaner. 
3. I had a problem with flake8 as above. Now the error is suppressed with `# noqa: F401`.

johannesring (approved):

PR #192, opened 2024-09-27
modify meshing script to save given meshing command

keiyamamo:
This PR introduces a feature to save command-line arguments to a file. 
This should ensure the reproducibility of the mesh.


johannesring:
You can already ensure the reproducibility of a mesh by using a config file and the `--config` option in `vasp-generate-mesh`, so I am not sure how useful this is.

johannesring (approved):
I am fine with merging this if you want to have this functionality.

keiyamamo:
I agree config does the job, but this could be convenient for situations where you want to test different parameters and keep track of the command over time. I will merge it but please let me know if this becomes annoying, we can then reverse the change anytime :)

PR #200, opened 2024-11-19
add a function to compute the minimum of Jacobian (displacement)

keiyamamo:
This PR adds a function to compute the minimum of Jacobian, which could be used to measure the quality of the mesh.
This can be used to help investigate the reason for divergence. 

Ref. Fig.6 in this paper: https://www.ljll.fr/~frey/papers/meshing/Wick%20T.,%20Fluid-structure%20interactions%20using%20different%20mesh%20motion%20techniques.pdf


johannesring (approved):
Looks good.

Issue #197, opened 2024-11-16
Assign different domain ID for different wall thickness


keiyamamo:
When using painted wall thickness, we also want to change the wall properties based on thickness. For example, thin wall region has soft wall. To do that, we need to assign different domain IDs so that `turtleFSI` can apply different solid properties for different regions.

keiyamamo:
Solved with #202

PR #202, opened 2024-11-27
Add support for thickness-based entity ID mapping in preprocessing

johannesring:
This PR (fixing issue #197) introduces functionality to map thickness values to mesh points and adjust cell entity IDs based on the mapped thickness. The new capabilities include:

* Thickness mapping: A `map_thickness_to_mesh` function to assign thickness values from a surface to mesh points.
* Entity ID updates: An `update_entity_ids_by_thickness` function to adjust cell entity IDs based on average point thickness within defined ranges.
* Command-line support: Added a `--thickness-to-entity-id-mapping` argument to define thickness-to-ID mappings directly via the command line.

This works when using both `variable` and `painted` thickness for the solid.


keiyamamo (approved):

keiyamamo:
It doesn’t have to be added to this PR, but I guess it would be nice to add a synthetic mesh with thickness array and add pytest for the painted surface.

johannesring:
I have added a test for this now. It is only for `variable` solid thickness (not `painted`).

PR #204, opened 2024-11-30
Minor update of spectrogram default parameters

keiyamamo:
This PR updates several default parameters for generating spectrograms. Also, I added `fsi_region` as a command-line argument, so that user flexibly modify its value. Finally, I added a check to see if specified `fsi_region` actually contains any nodes in it.


johannesring (approved):
Looks good!

PR #216, opened 2025-02-10
Removed “automated” from folder name

keiyamamo:
This PR rename the `automatedPreprocessing` & `automatedPostprocessing` to `pre_processing` & `post_processing`.
This avoids mixed usage of camelCase and snake_case and also simplify the navigation of folder.

There is no functionality-wise change.


johannesring (commented):
How about using `preprocessing` and `postprocessing` rather than `pre_processing` and `post_processing`? That would be more consistent with the names we already have, like `preprocessing_common`, `automated_preprocessing`, `postprocessing_fenics`, `postprocessing_h5py`, etc.

keiyamamo:
Yes, I thought about it. What do you think about the meshing script? Now it is named `automated_preprocessing.py`. Do you think it makes sense to change it to simply `preprocessing.py`?

johannesring:
Yes, `preprocessing.py` sounds good. Maybe `preprocessing_common.py` could be simply `common.py`?

johannesring:
Yes, in that case we should rename `simulation_common.py` to `common.py` as well (same for `postprocessing_common.py`). I am fine either way.

keiyamamo:
Thank you for your input! I would like to stick to `preprocessing_common.py` to avoid having files with same names. I will modify the PR shortly.

keiyamamo:
Hi @johannesring 

With latest commits, I modified folder name to be simply `preprocessing` and `postprocessing`. Could you please review it again?

johannesring (approved):

PR #225, opened 2025-05-03
updates on filtering visualization function

keiyamamo:
This PR includes a few updates on the filtering visualization functions. Previous implementation determined timestep based on parameters used for the FSI simulation. However, some post-processing is performed with stride. This update fixes this mismatch. Additionally, filtering of the stress is now included.

### DIFF
--- a/src/vasp/postprocessing/postprocessing_h5py/create_hi_pass_viz.py
+++ b/src/vasp/postprocessing/postprocessing_h5py/create_hi_pass_viz.py
@@ -20,7 +20,7 @@ from numpy import linalg as LA
 from tqdm import tqdm
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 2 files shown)


src/vasp/postprocessing/postprocessing_h5py/create_hi_pass_viz.py
```diff
@@ -564,7 +566,7 @@ def main():
 
     # In case of strain, modify the path to the solid mesh as the strain might be
     # computed for the solid domain only
-    if quantity == "strain":
+    if quantity == "strain" or "stress":
```
johannesring, line 569: There is a minor bug here since `stress` will always be `True`. Instead, it should be 
``` python
if quantity in {"strain", "stress"}:
```

